PR #34, opened 2018-06-01
Label match, events, binding/id create/del/edit, simplify sync loop

kkmsft:
Add ability to apply bindings based on pod labels.
Enable Eventing for binding apply and removal.
Add simpler Sync loop
Handle binding and id create/delete/edit.
Other misc fixes.


kkmsft:
cc @khenidak @rite2nikhil

PR #48, opened 2018-06-19
MIC permissions for identity created in another resource group

kkmsft:


README.md
```diff
@@ -88,6 +88,18 @@ Get the client id and resource id for the identity
 ```
 az identity create -g <resourcegroup> -n <idname>
 ```
+#### Providing required permissions for MIC
+
+This step is only required if you are using User assigned MSI.
+
+MIC uses the service principal credentials [stored within the the AKS](https://docs.microsoft.com/en-us/azure/aks/kubernetes-service-principal) cluster to access azure resources. This service principal needs to have Microsoft.ManagedIdentity/userAssignedIdentities/\*/assign/action permission on the identity for usage with User assigned MSI. If your identity is created in the same resource group as that of the AKS nodes (typically this resource group is prefixed with 'MC_' string and is automatically generatd when you create the cluster), you can skip the next steps.
```
xtellurian: generatd

PR #61, opened 2018-07-13
Add endpoint for host to get token

ritazh:
- [x] Return token and clientid to host
- [x] Get podns and podname from request header
- [x] Bump nmi image version
- [x] Doc new endpoint


pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
+func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.Request) {
+	hostIP := parseRemoteAddr(r.RemoteAddr)
+	 if hostIP != "127.0.0.1" {
+		msg := "request remote address is not from a host"
+		logger.Error(msg)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	 }
+
```
rite2nikhil: nit: spacing consistency in general

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	latency := time.Since(start)
 	logger.Infof("Status (%d) took %d ns", rw.statusCode, latency.Nanoseconds())
 }
+func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.Request) {
+	hostIP := parseRemoteAddr(r.RemoteAddr)
+	 if hostIP != "127.0.0.1" {
```
rite2nikhil: define as const

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		 return
+	 }
+	 rqClientID, rqResource := parseRequestClientIDAndResource(r)
+	 token, clientID, err := getTokenForMatchingID(logger, rqClientID, rqResource, podIDs)
+	 if err != nil {
+		 logger.Errorf("failed to get service principal token for pod:%s/%s, %+v", podns, podname, err)
+		 http.Error(w, err.Error(), http.StatusForbidden)
+		 return
+	 }
+	 //jsonToken, err := json.Marshal(*token)
```
rite2nikhil: remove comment if code not needed

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	 if hostIP != "127.0.0.1" {
+		msg := "request remote address is not from a host"
+		logger.Error(msg)
+		http.Error(w, msg, http.StatusInternalServerError)
+		return
+	 }
+
+	 podns, podname := parseRequstHeader(r)
+	 if podns == "" || podname == "" {
+		 logger.Errorf("Error getting podname and podns from request")
```
rite2nikhil: 1. small case first char to be consistent wit rest of the code

rite2nikhil: dont have to use "error" since the log is already marked as error, some like "missing pod name or namespaces in request header"

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		 http.Error(w, "Error getting podname and podns from request", http.StatusInternalServerError)
```
rite2nikhil: this is not internalserver error but a error on the client side of bad request, adding info in the error to how to send the pod name/namespace will make this more useful

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		 return
+	 }
+
+	 podIDs, err := s.KubeClient.ListPodIds(podns, podname)
+	 if err != nil || len(*podIDs) == 0 {
+		 msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s", podns, podname)
+		 logger.Errorf("%s, %+v", msg, err)
+		 http.Error(w, msg, http.StatusForbidden)
```
rite2nikhil: we can extract this as a function that can be used by both handlers ? to reduce code duplication

pkg/nmi/server/server.go
```diff
@@ -180,19 +230,28 @@ func getTokenForMatchingID(logger *log.Entry, rqClientID string, rqResource stri
...
+			return nil, clientID, fmt.Errorf("unsupported identity type %+v", idType)
 		}
 	}
 
 	// We have not yet returned, so pass up an error
-	return nil, fmt.Errorf("azureidentity is not configured for the pod")
+	return nil, "", fmt.Errorf("azureidentity is not configured for the pod")
+}
+
+func parseRequstHeader(r *http.Request) (podns string, podname string) {
```
rite2nikhil: misspent "Request" function name

pkg/nmi/server/server.go
```diff
@@ -180,19 +230,28 @@ func getTokenForMatchingID(logger *log.Entry, rqClientID string, rqResource stri
...
+	podns = r.Header.Get("podns")
```
rite2nikhil: add docs -- on handler definition and also on the readme doc in the NMI section

pkg/nmi/server/server.go
ritazh: Good idea... should i use `StatusUnprocessableEntity` 422 instead? 

Just FYI, in `msiHandler` we currently return `StatusInternalServerError` when podns and podname cannot be found. should we also return a diff status to be consistent? 
```
podns, podname, err := s.KubeClient.GetPodName(podIP)
	if err != nil {
		logger.Errorf("missing podname for podip:%s, %+v", podIP, err)
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
```

pkg/nmi/server/server.go
```diff
@@ -125,6 +132,49 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	 }
+
+	 podIDs, err := s.KubeClient.ListPodIds(podns, podname)
+	 if err != nil || len(*podIDs) == 0 {
+		 msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s", podns, podname)
+		 logger.Errorf("%s, %+v", msg, err)
+		 http.Error(w, msg, http.StatusForbidden)
+		 return
+	 }
+	 rqClientID, rqResource := parseRequestClientIDAndResource(r)
```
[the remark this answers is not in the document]
ritazh: Do you mean add retries for `getTokenForMatchingID` since that's where we need time to sync? Not the request for `resource`

pkg/nmi/server/server.go
```diff
@@ -134,46 +135,45 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	 podns, podname := parseRequstHeader(r)
-	 if podns == "" || podname == "" {
-		 logger.Errorf("Error getting podname and podns from request")
-		 http.Error(w, "Error getting podname and podns from request", http.StatusInternalServerError)
-		 return
-	 }
+	podns, podname := parseRequestHeader(r)
+	if podns == "" || podname == "" {
+		logger.Errorf("missing podname and podns from request")
+		http.Error(w, "missing 'podname' and 'podns' from request header", http.StatusUnprocessableEntity)
```
rite2nikhil: Returning StatusBadRequest (400) is more appropriate in this situation

pkg/nmi/server/server.go
rite2nikhil: this one is ok since the request came from the pod but server was not able to find the pod

pkg/nmi/server/server.go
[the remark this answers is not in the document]
rite2nikhil: i am asking to prevent the race when pods are being created and mic is trying to watch for pod events

PR #71, opened 2018-08-11
Json parse fixes and improve debuggability

kkmsft:
- Rename mic.go in cmd to main.go for better debuggability.

- Add more logging for error paths.

- Move to yaml.unmarshal instead of json.unmarshal to circumvent the json parsing
issues caused by values such as 'True' (with caps T instead of all small true)
in the value fields of azure.json. This is similar to the azure.go in upstream k8s
code (parseConfig function).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     docker:
       # specify the version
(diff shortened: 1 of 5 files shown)


ritazh:
LGTM

PR #82, opened 2018-09-10
nmi - filter on running pods - in some cases there can be non recycle…

rite2nikhil:
nmi - filter on running pods - in some cases there can be non recycled pod ips
- adding a filter to ip and running pods


pkg/k8s/client.go
```diff
@@ -33,51 +32,40 @@ type KubeClient struct {
...
-	}
-
 	if podip == "" {
 		return "", "", fmt.Errorf("podip is empty")
 	}
-
-	//podipFieldSel := fmt.Sprintf("status.podIp=%s", podip)
-	podList, err := c.ClientSet.CoreV1().Pods("").List(metav1.ListOptions{})
+	podList, err := c.ClientSet.CoreV1().Pods("").List(metav1.ListOptions{
+		FieldSelector: "status.podIP==" + podip + ",status.phase==Running",
```
ritazh: I think this should be filtering on pods in `Pending` and `Running` state.

rite2nikhil: The pods in pending state would not have an ip, so it should just be running state

kkmsft:
/lgtm for the code. 
Could we add a UT for this, please ?

PR #85, opened 2018-09-13
Update NMI version 

rite2nikhil:


ritazh:
`mcr.microsoft.com/k8s/aad-pod-identity/nmi:1.3` is not available yet

kkmsft:
The 1.3 nmi image should be available now. Removing the do-not-merge label.

The PR seems to having more changes than the issue at hand ?

Issue #94, opened 2018-10-01
Images should not be based on hyperkube


khenidak:
Please create an image based on alpine or similar small docker image. `hyperkube` is an overkill

@ritazh @kkmsft @rite2nikhil

khenidak:
This is fixed now by #95

PR #95, opened 2018-10-03
Switch to alpine as base image

ameukam:
As indicated #94, it's overkill to use hyperkube.


ameukam:
cc @kkmsft

khenidak:
/lgtm

@kkmsft Let us merge once the e2e webhook is enabled for the repo.

Issue #99, opened 2018-10-12
NMI - Retry GetPodName 


graniterocks:
We are experiencing a getting tokens from the nmi because the first call directly after pod startup fails: 
level=error msg="missing podname for podip:10.244.0.103, not found"

This error is similiar to what has been reported in issue #84 

After a short period of time the next call to the nmi works fine so I believe that a retry on the func (c *KubeClient) GetPodName(podip string)  would solve the problem.

PR #100, opened 2018-10-12
#99, added retry for GetPodName

graniterocks:


pkg/k8s/client.go
```diff
@@ -68,6 +75,21 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
 	return "", "", fmt.Errorf("match failed, ip:%s matching pods:%v", podip, podList)
 }
 
+func (c *KubeClient) getPodListWithTries(podip string, tries int) (*v1.PodList, error) {
```
kkmsft: Can we please take the sleep interval also as an option to this method ?

kkmsft: Can we please add UT which checks the retry ?

pkg/k8s/client.go
```diff
@@ -68,6 +75,21 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
+	podList, err := c.ClientSet.CoreV1().Pods("").List(metav1.ListOptions{
+		FieldSelector: "status.podIP==" + podip + ",status.phase==Running",
+	})
+
+	if err != nil || len(podList.Items) == 0 {
```
kkmsft: nit: Do we need to retry unnecessarily if its a fatal error. Since there is max retry time, this is not a fatal issue, but was wondering if there is a way for us to distinguish real errors vs expected errors ?

kkmsft:
Thank you !! @graniterocks for the contribution.
Overall /LGTM. Have placed some comments in the review.

PR #120, opened 2018-11-14
Add support for vmss

### LINKED ISSUE #63: Add VMSS support
When using a cluster with vmss agent pools created by acs-engine, getting the following error from nmi logs:

```
adal: Refresh request failed. Status Code = '400'. Response body: {\"error\":\"invalid_request\",\"error_description\":\"Identity not found\"}" req.method=GET req.path=/host/token/ req.remote=127.0.0.1
```
and this from the mic logs:

```
E0718 00:40:42.975094       1 mic.go:385] compute.VirtualMachinesClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="ResourceNotFound" Message="The Resource 'Microsoft.Compute/virtualMachines/k8s-agentpool-32078657-vmss000000' under resource group 'xxxx' was not found."
```

cpuguy83:
This adds support for assigning identities to a VMSS when the underlying
cluster is using VMSS.

Note that identities get assigned to the entire VMSS.

This introduces a dependency on being able to list nodes from k8s so we can get the provider I, which allows us to determine (per node) if it is VMSS and what the real name of the resource we need to assign to is. This has the additional benefit in that it should be more robust than relying on the node names from k8s to match node names in Azure.

Closes #63

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
+	"unicode"
 
 	config "github.com/Azure/aad-pod-identity/pkg/config"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
@@ -19,6 +21,7 @@ import (
 type Client struct {
 	ResourceGroupName string
 	VMClient          VMClientInt
+	VMSSClient        VMSSClientInt
 	ExtClient         compute.VirtualMachineExtensionsClient
 	Config            config.AzureConfig
 }
@@ -66,18 +69,28 @@ func NewCloudProvider(configFile string) (c *Client, e error) {
 	extClient.Authorizer = autorest.NewBearerAuthorizer(spt)
 	extClient.PollingDelay = 5 * time.Second
 
-	vmClient, err := NewVirtualMachinesClient(azureConfig, spt)
-	if err != nil {
-		glog.Errorf("Create VM Client error: %+v", err)
-		return nil, err
+	client := &Client{
+		ResourceGroupName: azureConfig.ResourceGroupName,
+		Config:            azureConfig,
+		ExtClient:         extClient,
 	}
 
-	return &Client{
-		ResourceGroupName: azureConfig.ResourceGroupName,
-		VMClient:          vmClient,
-		ExtClient:         extClient,
-		Config:            azureConfig,
-	}, nil
+	switch azureConfig.VMType {
+	case "vmss":
+		client.VMSSClient, err = NewVMSSClient(azureConfig, spt)
+		if err != nil {
+			glog.Errorf("Create VM Client error: %+v", err)
+			return nil, err
+		}
+	default:
+		client.VMClient, err = NewVirtualMachinesClient(azureConfig, spt)
+		if err != nil {
+			glog.Errorf("Create VM Client error: %+v", err)
+			return nil, err
+		}
+	}
+
+	return client, nil
 }
 
 func withInspection() autorest.PrepareDecorator {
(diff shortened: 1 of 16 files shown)


cpuguy83:
Main issue with this right now is the VM name is being passed through but we need to get at the VMSS name.

cpuguy83:
Removing WIP status.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -66,18 +70,24 @@ func NewCloudProvider(configFile string) (c *Client, e error) {
...
-	vmClient, err := NewVirtualMachinesClient(azureConfig, spt)
+	client := &Client{
+		ResourceGroupName: azureConfig.ResourceGroupName,
+		Config:            azureConfig,
+		ExtClient:         extClient,
+	}
+
+	client.VMSSClient, err = NewVMSSClient(azureConfig, spt)
+	if err != nil {
+		glog.Errorf("Create VM Client error: %+v", err)
```
ritazh: nit: `Create VMSS client error`

pkg/cloudprovider/vmss.go
```diff
@@ -0,0 +1,117 @@
...
+		glog.Error(err)
+		return vm, err
+	}
+	stats.Update(stats.CloudGet, time.Since(beginGetTime))
+	return vm, nil
+}
+
+// vmssIdentityHolder implements `IdentityHolder` for vmss resources.
+type vmssIdentityHolder struct {
+	vm *compute.VirtualMachineScaleSet
```
ritazh: nit: `vmss` instead of `vm`

pkg/cloudprovider/cloudprovider.go
```diff
@@ -91,109 +101,125 @@ func withInspection() autorest.PrepareDecorator {
...
+	glog.V(6).Infof("Get of %s completed in %s", node.Name, time.Since(timeStarted))
+
+	info := idH.IdentityInfo()
+	if info == nil {
+		info = idH.ResetIdentity()
+	}
+
+	info.AppendUserIdentity(userAssignedMSIID)
+
+	glog.Infof("ID: %s already found in vm identities", userAssignedMSIID)
```
cpuguy83: Oops, forgot to move this.

pkg/cloudprovider/cloudprovider.go
cpuguy83: Fixed

pkg/cloudprovider/identity.go
```diff
@@ -0,0 +1,92 @@
+package cloudprovider
+
+import (
+	"errors"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
+)
+
+// IdentityHolder represents a resource that contains an Identity object
+// This is used to be able to generically intract with multiple resource types (e.g. VirtualMachine and VittualMachineScaleSet)
```
kkmsft: nit: Vitt => Virt

pkg/cloudprovider/identity.go
```diff
@@ -0,0 +1,92 @@
...
+// This is needed because different Azure resource types (e.g. VirtualMachine and VirtualMachineScaleSet)
+// have different identity types.
+// This abstracts those differences.
+type IdentityInfo interface {
+	AppendUserIdentity(id string)
+	RemoveUserIdentity(id string) error
+}
+
+var (
+	errNotAssigned = errors.New("identity is not assigned to the resource")
```
kkmsft: Is it convention to add 'err' before errors. is that required ?

cpuguy83: It is convention. For error names, when defining an error type usually the end of the name has "Error", for variables it they tend to begin with "err"

pkg/cloudprovider/identity.go
cpuguy83: Fixed

pkg/cloudprovider/identity.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+}
+
+// appendUserIdentity provides a common implementation for adding a new identity
+// to an identity object.
+func appendUserIdentity(idType *compute.ResourceIdentityType, idList *[]string, assigned string) {
+	switch *idType {
+	case compute.ResourceIdentityTypeUserAssigned, compute.ResourceIdentityTypeSystemAssignedUserAssigned:
+		for _, id := range *idList {
```
kkmsft: Can you please add a comment here detailing what this loop is doing ?

pkg/cloudprovider/vm.go
```diff
@@ -69,3 +69,41 @@ func (c *VMClient) Get(rgName string, nodeName string) (ret compute.VirtualMachi
...
+	if err := filterUserIdentity(&i.info.Type, i.info.IdentityIds, id); err != nil {
+		return err
+	}
+	if i.info.Type == compute.ResourceIdentityTypeNone {
+		i.info.IdentityIds = nil
+	}
+	return nil
+}
+
+func (i *vmIdentityInfo) AppendUserIdentity(assigned string) {
```
kkmsft: what does assigned here mean ? is it the id to be assigned ?

pkg/cloudprovider/identity.go
```diff
@@ -0,0 +1,92 @@
...
+				return
+			}
+		}
+	case compute.ResourceIdentityTypeSystemAssigned:
+		*idType = compute.ResourceIdentityTypeSystemAssignedUserAssigned
+	default:
+		*idType = compute.ResourceIdentityTypeUserAssigned
+	}
+
+	*idList = append(*idList, assigned)
```
kkmsft: Was wondering if this abstracted enough here ? When we move to the dictionary based logic in the underlying sdk calls, would this have to be modified too ?

cpuguy83: I don't think we need to abstract here, this is just a common implementation for essentially the same operation.

If we change the API version then this would need to be updated accordingly, but I don't think we'd have mixed versions?

pkg/cloudprovider/identity.go
cpuguy83: Done

ritazh:
LGTM
I have also deployed this to a cluster with the demo app and keyvault-flexvol. Thanks @cpuguy83 🎉

omerlh:
Yay! Looking forward to test it out - how is the release process looks like? On which version will it be released?

PR #121, opened 2018-11-15
Small fix for e2e tests on master branch

chewong:
Had some modifications on identity validator and the e2e tests on master branch require the following small change to work. @kkmsft could you review this?


chewong:
Updated the identity validator docker image and will include the change in `identityvalidator:1.1`

PR #126, opened 2018-11-30
Additional e2e tests to ensure feature stability

chewong:
- Updated e2e test README.md
- Add `setup.sh` to ensure that necessary resources and role assignments are correct on Azure 
- Additional specs to further ensure feature stability
- Miscellaneous changes to improve code quality

@kkmsft feel free to review when you are available. Thanks!

PR #127, opened 2018-12-01
Fix image name for identity validator

pjbgf:


pjbgf:
This is already covered by #126

PR #128, opened 2018-12-02
Fix alpine migration

pjbgf:
* The alpine migration broke NMI and MIC, this PR fixes it by adding the package gcompat to both Dockerfiles. 
* Removes the iptables package from MIC.
* Fixes the image name for identity_validator.


pjbgf:
This fixes the issue #131. Do you mind having a look at this please @khenidak @ritazh ?

ritazh:
Will take a look today. Thanks @pjbgf!

pjbgf:
Fixed by #134

PR #130, opened 2018-12-03
Add xtables.lock mount to support iptables -w

### LINKED ISSUE #124: Mount /run/x-tables lockfile so that we don't interfere with other container touching iptables
See https://github.com/kubernetes/kubernetes/pull/46372 for reference and explainations

In short, when `iptables` is used from within a container, it should always use the `-w` flag (taken care of by the `iptables` go package used here) and the container should mount `/run/xtables.lock` to allow the lock to operate.

Implementation here : https://github.com/kubernetes/kubernetes/commit/0a1b7d94b441da381eb30c4890bb509771638f8c

/cc @khenidak

serbrech:
Fixes #124

### DIFF
--- a/charts/aad-pod-identity/templates/infra.yaml
+++ b/charts/aad-pod-identity/templates/infra.yaml
@@ -55,6 +55,7 @@ spec:
       volumes:
       - hostPath:
           path: /run/xtables.lock
+          type: FileOrCreate
         name: iptableslock
       containers:
       - name: nmi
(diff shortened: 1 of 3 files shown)


serbrech:
cc/ @khenidak

kkmsft:
In some cases, I am seeing races being reported on accessing the stables.lock file. Do we need to add to specify: FileOrCreate ?

serbrech:
Ah, I didn't know about this. It seems you are right : 
https://github.com/weaveworks/weave/pull/3353#issuecomment-406987249

I'll add it in, thanks

serbrech:
@kkmsft @khenidak rebased on master and pushed the `type: FileOrCreate` addition

kkmsft:
/lgtm
Request @khenidak to take a look too..

Issue #32, opened 2018-05-31
The Demo App app defaults to  some rando subscription/RG settings


ivanthelad:
attempting to run the demo app will result in a bunch of 403 errors as it tries to access a preset subscription, see output{1}, the demo app sets these values as defaults https://github.com/Azure/aad-pod-identity/blob/master/cmd/demo/main.go#L20

Easiest would be to update the deployment resource by setting a "command" and "argument" attribute to  override how the container is started and set the appropriate flags  there.
https://github.com/Azure/aad-pod-identity/blob/master/deploy/demo/deployment.yaml#L15

Readme, should be updated accordingly. 

{1} demoapp output 
`4-c2x24
time="2018-05-31T16:06:01Z" level=error msg="failed list all vm azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/c1089427-83d3-4286-9f35-5af546a6eb67/resourceGroups/MC_nbhatia-eu-01_eu-1_eastus/providers/Microsoft.Compute/virtualMachines?api-version=2017-12-01: StatusCode=403 -- Original Error: adal: Refresh request failed. Status Code = '403'. Response body: No AzureAssignedIdentity found for pod:default/demo-c6dc8f664-c2x24\n" podip=10.244.1.9 podname=demo-c6dc8f664-c2x24 podnamespace=demo-c6dc8f664-c2x24
time="2018-05-31T16:06:01Z" level=error msg="failed to refresh ServicePrincipalTokenFromMSI using the MSI VM extension, msiEndpoint(http://169.254.169.254/metadata/identity/oauth2/token)" podip=10.244.1.9 podname=demo-c6dc8f664-c2x24 podnamespace=demo-c6dc8f664-c2x24
time="2018-05-31T16:06:01Z" level=error msg="testMSIEndpoint failed, <nil>" podip=10.244.1.9 podname=demo-c6dc8f664-c2x24 podnamespace=demo-c6dc8f664-c2x24
[...]

PR #134, opened 2018-12-06
Build statically linked binaries

### LINKED ISSUE #131: Current master branch build is broken
Since the switch to alpine the Dockerfiles for both NMI and MIC are missing the package gcompat.

cpuguy83:
Ensures that these binaries work regardless of what libraries they were
linked to at build time... e.g. build on Ubuntu (with glibc) and run on
Alpine (musl libc).

The main thing here is that if you import the `net` or the `archive/tar` packages
go will automatically switch from the defaul statically linked build to a
dynamically linked one.

The other part of this, if you are cross compiling (e.g. when building from macOS or Windows.. makefile specifies GOOS=linux) you will also get a static build even without this change. So whether or not the alpine images work highly depends on what platform the binaries are built on.

Fixes #131

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ else
 	endif
 endif
 
-GO_BUILD_OPTIONS := -buildmode=${GO_BUILD_MODE} -ldflags "-s -X $(VERSION_VAR)=$(NMI_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE)"
+GO_BUILD_OPTIONS := -buildmode=${GO_BUILD_MODE} --tags "netgo osusergo" -ldflags "-s -X $(VERSION_VAR)=$(NMI_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE) -extldflags '-static'"
 E2E_TEST_OPTIONS := -count=1 -v
 
 # useful for other docker repos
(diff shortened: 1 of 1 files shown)


ritazh:
Getting the following error running on
Build OS: `4.15.0-1031-azure #32~16.04.1-Ubuntu`
Go Version: go1.11.1 linux/amd64
GCC: 5.4.0 20160609

```
make build
rm -rf bin/aad-pod-identity
rm -rf bin/aad-pod-identity/nmi
PKG_NAME=github.com/Azure/aad-pod-identity/cmd/nmi make bin/aad-pod-identity/nmi
make[1]: Entering directory '/go/src/github.com/Azure/aad-pod-identity'
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -buildmode=pie --tags "netgo osusergo" -ldflags "-s -X "github.com/Azure/aad-pod-identity"/version.Version=1.3 -X "github.com/Azure/aad-pod-identity"/version.GitCommit=$(git rev-parse --short HEAD) -X "github.com/Azure/aad-pod-identity"/version.BuildDate=$(date +%Y-%m-%d-%H:%M) -extldflags '-static'" -o "bin/aad-pod-identity/nmi" "github.com/Azure/aad-pod-identity/cmd/nmi"
# github.com/Azure/aad-pod-identity/cmd/nmi
loadinternal: cannot find runtime/cgo
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/5/crtbeginT.o: relocation R_X86_64_32 against `__TMC_END__' can not be used when making a shared object; recompile with -fPIC
/usr/lib/gcc/x86_64-linux-gnu/5/crtbeginT.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
```
On Mac, it's fine.

ritazh:
😞 I'm still getting the same error on two different ubuntu build machines with the latest commit. Maybe others can give this a try as well?

On a separate note, after updating Makefile (from master) to use `-buildmode=default` and `CGO_ENABLED=0`, the build works fine on both mac and ubuntu and runs successfully with alpine. While I do see the value of using `-buildmode=pie`, do we really need it since we aren't using any C stuff here.

pjbgf:
@cpuguy83 I tried the latest version of this PR on Ubuntu LTS 18 subsystem and on the base image: `circleci/golang:1.11.2`. In both cases it worked fine. For the latter, here's the result:
https://circleci.com/gh/pjbgf/aad-pod-identity/88

@ritazh I tried your suggestion of `-buildmode=default` and `CGO_ENABLED=0` and it worked fine on `circleci/golang:1.11.2`:
https://circleci.com/gh/pjbgf/aad-pod-identity/92

How can we move forwards with this and get the change merged?

kkmsft:
@pjbgf : Thank you for raising the issue and validating the fix.

cpuguy83:
Let's make sure this is good for @ritazh 

In testing, it does seem like we can `builmode=default` and `CGO_ENABLED=0`, this may be ideal rather than the rest of the work-arounds to ensure that `buildmode=pie` works.

PR #135, opened 2018-12-07
Add Jenkinsfile for building and pushing images to registry

chewong:
Also converted  the demo Dockerfile to an alpine-based image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,10 +18,9 @@ pipeline {
 
         stage('Build images and push to registry') {
             steps {
-                sh """
-                go get github.com/Azure/aad-pod-identity || true
-                cd /root/go/src/github.com/Azure/aad-pod-identity && /root/go/bin/dep ensure && make build && make image && make push
-                """
+                sh 'go get -u github.com/golang/dep/cmd/dep'
+                sh 'go get github.com/Azure/aad-pod-identity || true'
+                sh 'cd /root/go/src/github.com/Azure/aad-pod-identity && /root/go/bin/dep ensure && make build && make image && make push'
             }
         }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,7 @@ pipeline {
     stages {
         stage('Setup building environment') {
             steps {
+                // musl-dev is a dependency of go
                 sh 'apk upgrade && apk update && apk add --no-cache go make musl-dev docker'
                 withCredentials([azureServicePrincipal("${SP_CREDENTIAL}")]) {
                     sh 'az login --service-principal -u "$AZURE_CLIENT_ID" -p "$AZURE_CLIENT_SECRET" -t "$AZURE_TENANT_ID"'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,28 @@
+pipeline {
+    agent {
+        docker {
+            image 'microsoft/azure-cli'
+            args '-u root:root -v /var/run/docker.sock:/var/run/docker.sock'
+        }
+    }
+    stages {
+        stage('Setup building environment') {
+            steps {
+                sh 'apk upgrade && apk update && apk add --no-cache go make musl-dev docker'
+                withCredentials([azureServicePrincipal("${SP_CREDENTIAL}")]) {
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,27 @@
...
+    agent {
+        docker {
+            image 'microsoft/azure-cli'
+            args '-u root:root -v /var/run/docker.sock:/var/run/docker.sock'
+        }
+    }
+    stages {
+        stage('Setup building environment') {
+            steps {
+                sh 'apk upgrade && apk update && apk add --no-cache go make musl-dev docker'
```
kkmsft: Can you please add a comment here indicating what the building of musl-dev means and also where this would get executed ?

Makefile
```diff
@@ -33,10 +33,10 @@ E2E_TEST_OPTIONS := -count=1 -v
 
 # useful for other docker repos
 REGISTRY ?= mcr.microsoft.com/k8s/aad-pod-identity
-NMI_IMAGE_NAME := $(REGISTRY)/$(NMI_BINARY_NAME)
-MIC_IMAGE_NAME := $(REGISTRY)/$(MIC_BINARY_NAME)
-DEMO_IMAGE_NAME := $(REGISTRY)/$(DEMO_BINARY_NAME)
-IDENTITY_VALIDATOR_IMAGE_NAME := $(REGISTRY)/$(IDENTITY_VALIDATOR_BINARY_NAME)
+NMI_IMAGE_NAME := $(REGISTRY)/$(NMI_BINARY_NAME):$(NMI_VERSION)
```
kkmsft: Can you please change this part to make it possible to read the image names from environment ?

kkmsft: Can we also change it so that we can accept env vars for the versions too ?

Jenkinsfile
```diff
@@ -0,0 +1,27 @@
...
+                }
+                sh 'az acr login -n "$REGISTRY_NAME"'
+            }
+        }
+
+        stage('Build images and push to registry') {
+            steps {
+                sh 'go get -u github.com/golang/dep/cmd/dep'
+                sh 'go get github.com/Azure/aad-pod-identity || true'
+                sh 'cd /root/go/src/github.com/Azure/aad-pod-identity && /root/go/bin/dep ensure && make build && make image && make push'
```
kkmsft: Can we please make the build version also a possible parameter from Jenkins ?

Makefile
```diff
@@ -76,33 +76,37 @@ deepcopy-gen:
...
+	docker push $(NMI_IMAGE_NAME)
 
 push-mic:
-	docker push $(MIC_IMAGE_NAME):$(MIC_VERSION)
+	docker pull $(MIC_IMAGE_NAME); if [[ $$? -eq 0 ]]; then echo "$(MIC_IMAGE_NAME) already exists" && exit 1; fi
+	docker push $(MIC_IMAGE_NAME)
 
 push-demo:
-	docker push $(DEMO_IMAGE_NAME):$(DEMO_VERSION)
+	docker pull $(DEMO_IMAGE_NAME); if [[ $$? -eq 0 ]]; then echo "$(DEMO_IMAGE_NAME) already exists" && exit 1; fi
```
kkmsft: Can we please use 'docker images' instead of pull so that we don't have to bring the image to the local system, instead just check if it's present.

Makefile
```diff
@@ -32,11 +32,13 @@ GO_BUILD_OPTIONS := -buildmode=${GO_BUILD_MODE} --tags "netgo osusergo" -ldflags
...
 # useful for other docker repos
-REGISTRY ?= mcr.microsoft.com/k8s/aad-pod-identity
-NMI_IMAGE_NAME := $(REGISTRY)/$(NMI_BINARY_NAME)
-MIC_IMAGE_NAME := $(REGISTRY)/$(MIC_BINARY_NAME)
-DEMO_IMAGE_NAME := $(REGISTRY)/$(DEMO_BINARY_NAME)
-IDENTITY_VALIDATOR_IMAGE_NAME := $(REGISTRY)/$(IDENTITY_VALIDATOR_BINARY_NAME)
+REGISTRY_NAME ?= upstreamk8sci
+REGISTRY ?= upstreamk8sci.azurecr.io
+REPO_PREFIX ?= k8s/aad-pod-identity
+NMI_IMAGE_NAME ?= $(REPO_PREFIX)/$(NMI_BINARY_NAME):$(NMI_VERSION)
```
kkmsft: Can you please remove the _NAME since this has both name and version.

Issue #115, opened 2018-11-12
Proposed Behavior Change: Identity/Namespaces/Assignment


khenidak:
Ref: #106 #45 and other

Today, Identities are hardcoded to default namespace. and pods are matched using selector irrespective of `pod.namespace`. The is causing the following:
1. Users would like to compartmentalize application resources at the namespace level. This includes the identity.
2. This model currently does not allow matching `identity.namespace == pod.namespace` 

**Proposed change**:

1. Remove the hardcoded selector in https://github.com/Azure/aad-pod-identity/blob/master/pkg/crd/crd.go#L128 (along with any reference to default namespace in `listXXX` calls)
2. Add annotation to `azureIdentity` to force matching to current namespace only. i.e `true = (pod.namespace == azureIdentity.namespace)` When the annotation is not found then we match in any namespace.

@kkmsft @rite2nikhil @ritazh Thoughts?

rite2nikhil:
The default is to match on any namespace, we will be backward compatible /lgtm

brantb:
Out of curiosity, what is the motivation behind selecting pods based on the label `aadpodidbinding` instead of allowing arbitrary label selectors [as described by the k8s documentation?](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) Typically a namespaced resource will typically reference either (a) other resources in the same namespace, or (b) a non-namespaced resource like a `ClusterRole` or a `PersistentVolume`.

I'd like to suggest the following behavior:

1. Allow `AzureIdentity` and `AzureIdentityBinding` to be placed in any namespace.
2. [breaking] Change the spec of `AzureIdentityBinding` to reference pods in the current namespace using an arbitrary selector, e.g.
```yaml
apiVersion: aadpodidentity.k8s.io/v1
kind: AzureIdentityBinding
metadata:
  name: demo-azure-id-binding
  namespace: my-namespace
spec:
  AzureIdentity: my-azure-identity-name
  selector:
    matchLabels:
      app: myapplication # example
      environment: dev   # example
```
3. [breaking] When an `AzureIdentityBinding` references an `AzureIdentity` by name in `spec.AzureIdentity`, match only in the same namespace. (Alternate but also breaking idea: Make `AzureIdentity` a cluster-wide resource so that they do not reside in any namespace)

I completely understand if you don't want to make breaking changes at this time, but this would bring aad-pod-identity more in line with how I've seen many other Kubernetes applications operate.

[...]

PR #140, opened 2019-01-07
namespaced identities

khenidak:
This PR implements https://github.com/Azure/aad-pod-identity/issues/115 and https://github.com/Azure/kubernetes-keyvault-flexvol/issues/48

Also included: 
1. Fix leaky abstraction in mic implementation.
2. Modified and updated error messages.

> While there is a proposal to switch to selectors, i want to unblock this request first then figure out how we can do selectors.

@kkmsft @rite2nikhil @ritazh 

CC: @pjbgf


pkg/mic/mictest/mic_test.go
```diff
@@ -80,48 +80,51 @@ func TestSimpleMICClient(t *testing.T) {
...
 	if listAssignedIDs != nil {
 		for _, assignedID := range *listAssignedIDs {
 			if assignedID.Spec.Pod == "test-pod" && assignedID.Spec.PodNamespace == "default" && assignedID.Spec.NodeName == "test-node" &&
 				assignedID.Spec.AzureBindingRef.Name == "testbinding" && assignedID.Spec.AzureIdentityRef.Name == "test-id" {
-				testPass = evtRecorder.Validate(&LastEvent{Type: "Normal", Reason: "binding applied",
-					Message: "Binding testbinding applied on node test-node for pod test-pod-default-test-id"})
-				if !testPass {
-					panic("event mismatch")
-				}
+				/*
```
rite2nikhil: need this ?

pkg/mic/mictest/mic_test.go
```diff
@@ -80,48 +80,51 @@ func TestSimpleMICClient(t *testing.T) {
...
-		panic("list assigned failed")
-	}
-	testPass = evtRecorder.Validate(&LastEvent{Type: "Normal", Reason: "binding removed",
-		Message: "Binding testbinding removed from node test-node for pod test-pod"})
-
-	if !testPass {
-		panic("event mismatch")
+		t.Errorf("list assigned failed")
 	}
+	/*
```
rite2nikhil: comment needed ?

Makefile
```diff
@@ -28,7 +28,7 @@ else
 	endif
 endif
 
-GO_BUILD_OPTIONS := -buildmode=${GO_BUILD_MODE} --tags "netgo osusergo" -ldflags "-s -X $(VERSION_VAR)=$(NMI_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE) -extldflags '-static'"
+GO_BUILD_OPTIONS := -buildmode=${GO_BUILD_MODE} --tags "netgo osusergo"  -ldflags "-s -X $(VERSION_VAR)=$(NMI_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE) -extldflags '-static'"
```
rite2nikhil: why add a space ?

pkg/mic/mictest/mic_test.go
khenidak: I am slowly changing the way we test. This part is now un needed because it was an abstraction exposed by mic. I will remove the commented code as finalize the PR

cmd/mic/main.go
```diff
@@ -10,14 +11,16 @@ import (
...
+	kubeconfig      string
+	cloudconfig     string
+	forceNamespaced bool
 )
 
 func main() {
 	defer glog.Flush()
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
+	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces mic to namespace identities, binding, and assignment")
```
rite2nikhil: nit: in the context of mic "mic to" is redundant, can be modified to
"Forces namespace identities, binding, and assignment"

cmd/nmi/main.go
```diff
@@ -35,7 +36,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("%+v", err)
 	}
-	s := server.NewServer()
+	s := server.NewServer(*forceNamespaced || "true" == os.Getenv("FORCENAMESPACED"))
```
rite2nikhil: nit: its better to update *forcedNamespaced variable with this check and not doing the check in the NewServer call

pkg/crd/crdtest/crd_test_util.go
```diff
@@ -35,25 +35,36 @@ func (c *TestCrdClient) CreateCrdWatchers(eventCh chan aadpodid.EventType) (err
...
 
-func (c *TestCrdClient) RemoveAssignedIdentity(name string) error {
-	delete(c.assignedIDMap, name)
+func (c *TestCrdClient) RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureAssignedIdentity) error {
+	delete(c.assignedIDMap, assignedIdentity.Name)
 	return nil
 }
 
-func (c *TestCrdClient) CreateAssignedIdentity(name string, binding *aadpodid.AzureIdentityBinding, id *aadpodid.AzureIdentity, podName string, podNameSpace string, nodeName string) error {
+// This function is not used currently
```
rite2nikhil: is thic comment relevant ?

pkg/mic/mic.go
```diff
@@ -194,30 +160,39 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
 			}
 
 			for _, binding := range matchedBindings {
 				glog.V(5).Infof("Looking up id map: %v", binding.Spec.AzureIdentity)
 				if azureID, idPresent := idMap[binding.Spec.AzureIdentity]; idPresent {
-					glog.V(5).Infof("Id %s got for assigning", azureID.Name)
-					assignedID, err := c.MakeAssignedIDs(&azureID, &binding, pod.Name, pod.Namespace, pod.Spec.NodeName)
+					if c.IsNamespaced || aadpodid.IsNamespacedIdenity(&azureID) { // working in Namespaced mode or this specific identity is namespaced
+						if !(azureID.Namespace == binding.Namespace && binding.Namespace == pod.Namespace) { // They have to match all
+							glog.V(2).Infof("identity %s/%s was matched via binding %s/%s to %s/%s but namespaced identity is enforced, so it will be ignored",
```
rite2nikhil: nit: Comment in new line (top of the line) to be consistent

rite2nikhil: should this we warning ?

pkg/mic/mic.go
```diff
@@ -462,6 +441,32 @@ func (c *Client) RemoveAssignedIDsWithDeps(assignedID *aadpodid.AzureAssignedIde
...
+	return m, nil
+}
+
+func (c *Client) checkIfInUse(checkAssignedID aadpodid.AzureAssignedIdentity, arr []aadpodid.AzureAssignedIdentity) bool {
+	for _, assignedID := range arr {
+		checkID := checkAssignedID.Spec.AzureIdentityRef
+		id := assignedID.Spec.AzureIdentityRef
+		// If they have the same client id, reside on the same node but the pod name is different, then the
+		// assigned id is in use.
+		//This is applicable only for user assigned MSI since that is node specific. Ignore other cases.
```
rite2nikhil: nit: add space after //  in comment

pkg/mic/mictest/mic_test.go
```diff
@@ -25,36 +25,36 @@ func TestMapMICClient(t *testing.T) {
...
 	idList = append(idList, *id)
-
-	idMap, _ := micClient.ConvertIDListToMap(&idList)
-
-	name := "test-azure-identity"
-	count := 3
-	if azureID, idPresent := idMap[name]; idPresent {
-		if azureID.Name != name {
-			panic("id map id value mismatch")
+	/*
```
rite2nikhil: commented test ?

pkg/nmi/server/server.go
```diff
@@ -155,6 +158,21 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
...
 		return
 	}
+
+	// filter out if we are in namesoaced mode
+	filterPodIdentities := []aadpodid.AzureIdentity{}
+	for _, val := range *(podIDs) {
+		if s.IsNamespaced || aadpodid.IsNamespacedIdenity(&val) { // namespaced mode
+			if val.Namespace == podns { // matched namespace
+				filterPodIdentities = append(filterPodIdentities, val)
+			} else { // unmatcjed namespaced
```
rite2nikhil: spell check and comment in new line

pkg/nmi/server/server.go
```diff
@@ -155,6 +158,21 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
 		http.Error(w, msg, http.StatusForbidden)
 		return
 	}
+
+	// filter out if we are in namesoaced mode
+	filterPodIdentities := []aadpodid.AzureIdentity{}
+	for _, val := range *(podIDs) {
+		if s.IsNamespaced || aadpodid.IsNamespacedIdenity(&val) { // namespaced mode
```
rite2nikhil: optional: can reduce the branching by using if (condition) { assignment; continue} pattern

pkg/apis/aadpodidentity/v1/utils.go
```diff
@@ -0,0 +1,10 @@
+package v1
+
+func IsNamespacedIdenity(azureId *AzureIdentity) bool {
```
ritazh: typo: `IsNamespacedIdenity` ==> `IsNamespacedIdentity`

README.md
```diff
@@ -124,6 +124,13 @@ spec:
 kubectl create -f aadpodidentity.yaml
 ```
 
+#### Understanding Namespaced identities
+The system will match `pod` to `identity` across namespaces by default. This behavior can be modified to match to pods within the namespace that holds `AzureIdentity` by
+
+1. On Azure Identity bases, by addinging `aadpodidentity.k8s.io/Behavior: namespaced` annotation (You have to add the annotation on each `AzureIdentity` you want to apply this behavior on).
+2. Default namespaced behavior on all identities by adding `--forceNamespaced=true` argument on the command line  or declare `FORCENAMESPACED=true` environment variable (for both `nmi` and `mic`).
```
ritazh: We should document performing 2 can override configurations set in 1 given this `if s.IsNamespaced || aadpodid.IsNamespacedIdenity(&val)` in mic.go

pkg/crd/crd.go
```diff
@@ -98,7 +98,7 @@ func newRestClient(config *rest.Config) (r *rest.RESTClient, err error) {
 
 func newBindingWatcher(r *rest.RESTClient, eventCh chan aadpodid.EventType) (cache.SharedInformer, error) {
 	azBindingWatcher := cache.NewSharedInformer(
-		cache.NewListWatchFromClient(r, aadpodid.AzureIDBindingResource, "default", fields.Everything()),
+		cache.NewListWatchFromClient(r, aadpodid.AzureIDBindingResource, v1.NamespaceAll, fields.Everything()),
```
cpuguy83: Is this a breaking change?
Before we were only watching `default` now all namespaces?

pkg/crd/crdtest/crd_test_util.go
cpuguy83: If it's not being called we should remove. Can always bring it back from history if needed.

pkg/mic/mic.go
```diff
@@ -410,24 +385,28 @@ func (c *Client) MakeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
 			AvailableReplicas: 1,
 		},
 	}
+	// if we are in namespaced mode (or az identity is namespaced)
```
cpuguy83: comment seems redundant?

cpuguy83: Also do we need the if check? Would `azID.Namespace` be set to what we want regardless of it actually being namespaced?

pkg/mic/mic.go
```diff
@@ -462,6 +441,32 @@ func (c *Client) RemoveAssignedIDsWithDeps(assignedID *aadpodid.AzureAssignedIde
 	return nil
 }
 
-func (c *Client) GetAssignedIDName(podName string, podNameSpace string, idName string) string {
+func (c *Client) getAssignedIDName(podName string, podNameSpace string, idName string) string {
 	return podName + "-" + podNameSpace + "-" + idName
 }
+
+func (c *Client) convertIDListToMap(arr *[]aadpodid.AzureIdentity) (m map[string]aadpodid.AzureIdentity, err error) {
+	m = make(map[string]aadpodid.AzureIdentity)
```
cpuguy83: specify map size to `len(arr)` to avoid multiple allocs below.

pkg/crd/crd.go
khenidak: With the new changes it will look exactly the same. For unnamespaced identities. everything will be created on the default namespace

khenidak:
@kkmsft I refactored all the unit tests. Each package now is tested independently without the cross reference we had before. I have also put back the majority of the previously commented tests, I will be adding the rest shortly.

pkg/mic/mictest/mic_test.go
khenidak: Yeah was in the process of refactoring. Now tests are back (except the ones related to events)

README.md
khenidak: I am not sure i understand this one, can you please elaborate?

pkg/mic/mic.go
khenidak: good catch.. thank you

README.md
ritazh: If `--forceNamespaced=true` is set as described in 2, then the namespaced behavior is set on all identities, regardless what annotation is used on the individual AzureIdentity. Hence setting 2 can override configurations set in 1. This behavior should be called out.

pjbgf: @khenidak @ritazh  I understand the motivation for avoiding breaking changes, however, from a security standpoint, it would be better to have `forceNamespaced=false` by default. That would this application a bit more "secure by default", which would be easier for people to use it without needing to know much about the project. 

If the priority of not breaking compatibility is higher, we probably should emphasise in the documentation the potential lateral movement within Azure Control Plane, specially for cases of clusters hosting multiple applications/used by multiple teams.

cpuguy83: A hard break may be desirable but is dangerous from an end-user perspective. We should probably follow a formal deprecation policy and communicate after X time the default will be flipped to the more secure model.

ritazh: > from a security standpoint, it would be better to have forceNamespaced=false by default. That would this application a bit more "secure by default",

@pjbgf Perhaps I'm misunderstanding your point here but wouldnt `forceNamespaced=false` make it less secure since we are assigning the identity to any binding that match regardless of the namespace?
https://github.com/Azure/aad-pod-identity/blob/6ac452027c9246ec057a7b66c551c1e9c6e8c2ab/pkg/mic/mic.go#L185-L192

khenidak: `forceNamespaced=false` is the default. it is an opt in feature to maintain back combat. Please keep in mind that namespaced identities/binding/assignment is not a security feature per se, it is more of an organizational feature.

I want us to maintain back combat (unless we `absolutely` have to). I hate to ask users to modify anything (which may include an app downtime) they have already working.

doug-fish: I'm a likely consumer of this capability. I don't have any existing deployments (so I don't personally care about backward compatibility). I don't perceive a problem with requiring me to specify forceNamespaced=true to get the behavior I want.

I perceive this to be a security feature; it makes me uneasy if you think it's just an organizational feature. I'm standing up AKS with AD integration (so I know who the users are), and using Kubernetes RBAC to restrict users to only work in authorized namespaces and with authorized object types. I perceive this integration to be a useful way for each namespace to be associated with an azure service principal so that user (and only that user) can securely connection to Azure resources outside the cluster that have potentially sensitive data.

Will using Azure pod identity and keyvault-flexvol introduce security concerns?

pjbgf: @ritazh You are right, I meant to say `forceNamespaced=true`. I just edited that to make sense. 
I also highlighted [here](https://medium.com/@pjbgf/using-aad-pod-identity-in-your-azure-kubernetes-clusters-what-to-watch-out-for-73d5d73960f) three scenarios (1, 2 and 3) in which having `forceNamespaced=false` makes it easier for non-admins to do lateral movements between the cluster control plane and the azure control plane.

Issue #141, opened 2019-01-09
Incorrectly appending create list to the delete list


Pothulapati:
Hey,

In the `SplitAzureAssignedIDs` method, if the condition is `in(old) && !in(new)`  then we have delete. So, We should be appending the id to the `delete` list but it is incorrectly being appended to the `create` list as shown here.
https://github.com/Azure/aad-pod-identity/blob/2540a613757050f326755062e6d5a3682d9f8371/pkg/mic/mic.go#L388

If we see the other condition, i.e `!in(old) && in(new)`, then we have to create the id. So we are appending them rightly to the `create`  list as shown here.
https://github.com/Azure/aad-pod-identity/blob/2540a613757050f326755062e6d5a3682d9f8371/pkg/mic/mic.go#L365

Did I understand something wrong?

kkmsft:
Thank you @Pothulapati. 

The issue manifests when there are delete and add in the same cycle. Because we are running delete cycles first and then performing create cycles, the effect of create list adding itself to delete gets nullified there because create eventually gets done, right after the delete. Nevertheless this causes an unnecessary deletion and addition to the underlying vm identity. As for deletions take effect one at a time because only one 'real' to-be-deleted one gets appended in one cycle. Eventually the others also get deleted in subsequent cycles.

Working on the fix and test to repro any side effects of this.

kkmsft:
Fix merged. Closing the bug.

PR #142, opened 2019-01-09
Fixes the issue #141 raised for delete list appended with create list

kkmsft:
The issue manifests when there are delete and add in the same cycle. Because we are running delete cycles first and then performing create cycles, the effect of create list adding itself to delete gets nullified there because create eventually gets done, right after the delete. Nevertheless this causes an unnecessary deletion and addition to the underlying vm identity. As for deletions take effect one at a time because only one 'real' to-be-deleted one gets appended in one cycle. Eventually the others also get deleted in subsequent cycles.

Inorder to repro the issue, the test adds two pods in one cycle, and then deletes both and and adds one more pod in the next cycle. With the bug - only one assigned id gets deleted in one cycle and the second one gets deleted in the second cycle. With the test both gets deleted and the assigned identity for the new pod gets added in one cycle.

### DIFF
--- a/pkg/crd/crd_test.go
+++ b/pkg/crd/crd_test.go
@@ -1,7 +1,7 @@
 package crd
 
 import (
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
@@ -42,28 +42,7 @@ func (c *TestCrdClient) RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureA
 // This function is not used currently
 // TODO: consider remove
 func (c *TestCrdClient) CreateAssignedIdentity(assignedIdentity *aadpodid.AzureAssignedIdentity) error {
-	assignedID := &aadpodid.AzureAssignedIdentity{
-		ObjectMeta: v1.ObjectMeta{
-			Name:      "some-name",
-			Namespace: "default",
-		},
-		Spec: aadpodid.AzureAssignedIdentitySpec{
-			Pod:          "test-pod",
-			PodNamespace: "defaut",
-			NodeName:     "test-node",
-			AzureBindingRef: &aadpodid.AzureIdentityBinding{
-				ObjectMeta: v1.ObjectMeta{
-					Name: "testbinding",
-				},
-			},
-			AzureIdentityRef: &aadpodid.AzureIdentity{
-				ObjectMeta: v1.ObjectMeta{
-					Name: "test-id",
-				},
-			},
-		},
-	}
-	c.assignedIDMap["some-name"] = assignedID
+	c.assignedIDMap[assignedIdentity.Name] = assignedIdentity
 	return nil
 }
 
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -360,9 +360,9 @@ func (c *Client) splitAzureAssignedIDs(old *[]aadpodid.AzureAssignedIdentity, ne
 		}
 		if !idMatch {
 			glog.V(5).Infof("ok: %v, Delete added: %s", idMatch, oldAssignedID.Name)
-			// We are done checking that this new id is not present in the old
-			// list. So we will add it to the create list.
-			delete = append(create, oldAssignedID)
+			// We are done checking that this old id is not present in the new
+			// list. So we will add it to the delete list.
+			delete = append(delete, oldAssignedID)
 		}
 	}
 	stats.Put(stats.FindAssignedIDDel, time.Since(begin))
--- a/pkg/mic/mic_test.go
+++ b/pkg/mic/mic_test.go
@@ -16,7 +16,7 @@ import (
 	cp "github.com/Azure/aad-pod-identity/pkg/cloudprovider"
 	api "k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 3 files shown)


cpuguy83:
@kkmsft Is this ready for review?

kkmsft:
yes, please go ahead @cpuguy83. I need to add a simpler UT which just calls that specific method, but the current code itself is ready and waiting for review :-)

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+	podClient.GetPods()
+
+	crdClient.CreateId("test-id4", aadpodid.UserAssignedMSI, "test-user-msi-resourceid", "test-user-msi-clientid", nil, "", "", "")
+	crdClient.CreateBinding("testbinding4", "test-id4", "test-select4")
+	podClient.AddPod("test-pod4", "default", "test-node2", "test-select4")
+	podClient.GetPods()
+
+	eventCh <- aadpodid.PodCreated
+	go micClient.Sync(exit)
+	time.Sleep(5 * time.Second)
```
cpuguy83: Is there a better way to do this?
Sleep like this is racey and slow.

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+
+	//Add a new pod, with different id and binding on the same node.
+	crdClient.CreateId("test-id3", aadpodid.UserAssignedMSI, "test-user-msi-resourceid", "test-user-msi-clientid", nil, "", "", "")
+	crdClient.CreateBinding("testbinding3", "test-id3", "test-select3")
+	podClient.AddPod("test-pod3", "default", "test-node2", "test-select3")
+	podClient.GetPods()
+
+	eventCh <- aadpodid.PodCreated
+	go micClient.Sync(exit)
+	time.Sleep(5 * time.Second)
```
cpuguy83: Same

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+		glog.Error(err)
+		panic("list assigned failed")
+	}
+
+	expectedLen = 1
+	gotLen = len(*listAssignedIDs)
+	//One id should be left around. Rest should be removed
+	if gotLen != expectedLen {
+		glog.Errorf("Expected len: %d. Got: %d", expectedLen, gotLen)
+		panic("Add and delete id at same time mismatch")
```
cpuguy83: `t.Fatal()`

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+	} else {
+		gotID := (*listAssignedIDs)[0].Name
+		expectedID := "test-pod3-default-test-id3"
+		if gotID != expectedID {
+			glog.Errorf("Expected %s. Got: %s", expectedID, gotID)
+			panic("Add and delete id at same time. Found wrong id")
```
cpuguy83: `t.Fatal()`

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+
+	//Delete the pod
+	podClient.DeletePod("test-pod2", "default")
+	podClient.DeletePod("test-pod4", "default")
+
+	//Add a new pod, with different id and binding on the same node.
+	crdClient.CreateId("test-id3", aadpodid.UserAssignedMSI, "test-user-msi-resourceid", "test-user-msi-clientid", nil, "", "", "")
+	crdClient.CreateBinding("testbinding3", "test-id3", "test-select3")
+	podClient.AddPod("test-pod3", "default", "test-node2", "test-select3")
+	podClient.GetPods()
```
cpuguy83: What is this doing?

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+	micClient := NewMICClient(eventCh, cloudClient, crdClient, podClient, nodeClient, &evtRecorder)
+
+	// Test to add and delete at the same time.
+	// Add a pod, identity and binding.
+	crdClient.CreateId("test-id2", aadpodid.UserAssignedMSI, "test-user-msi-resourceid", "test-user-msi-clientid", nil, "", "", "")
+	crdClient.CreateBinding("testbinding2", "test-id2", "test-select2")
+
+	nodeClient.AddNode("test-node2")
+	podClient.AddPod("test-pod2", "default", "test-node2", "test-select2")
+	podClient.GetPods()
```
cpuguy83: ?

pkg/mic/mictest/mic_test.go
```diff
@@ -159,3 +159,83 @@ func TestSimpleMICClient(t *testing.T) {
...
+
+	crdClient.CreateId("test-id4", aadpodid.UserAssignedMSI, "test-user-msi-resourceid", "test-user-msi-clientid", nil, "", "", "")
+	crdClient.CreateBinding("testbinding4", "test-id4", "test-select4")
+	podClient.AddPod("test-pod4", "default", "test-node2", "test-select4")
+	podClient.GetPods()
```
cpuguy83: ?

pkg/mic/mictest/mic_test.go
kkmsft: Adding a new pod, id and binding.

kkmsft: oh GetPods, that is called without any effects. Commenting it out.

Issue #146, opened 2019-01-20
MIC deployment is not restricted to Linux nodes and the image does not support windows


ankrause:
In a hybrid or windows-based cluster, the current deployment specs allow the MIC pod to be scheduled on windows nodes, despite not supporting windows.

[Deployment](https://github.com/Azure/aad-pod-identity/blob/master/deploy/infra/deployment.yaml)
[Deployment - RBAC](https://github.com/Azure/aad-pod-identity/blob/master/deploy/infra/deployment-rbac.yaml)

```
Failed to pull image "mcr.microsoft.com/k8s/aad-pod-identity/mic:1.2": rpc error: code = Unknown desc = image operating system "linux" cannot be used on this platform
```

NMI though is already restricted to Linux nodes.

PR #149, opened 2019-01-26
Add linux node selector to MIC and demo deployments

ankrause:
Address #146 by adding the linux node selector to all deployments of MIC.

Also added the selector to the demo deployment as the current image is also linux-specific.


kkmsft:
Thank you for your contribution, @ankrause.

PR #151, opened 2019-02-01
update containers' images' versions

khenidak:
Update charts and deployment spec to match most recent versions

CC @kkmsft

### DIFF
--- a/charts/aad-pod-identity/templates/infra.yaml
+++ b/charts/aad-pod-identity/templates/infra.yaml
@@ -59,9 +59,10 @@ spec:
         name: iptableslock
       containers:
       - name: nmi
-        image: "mcr.microsoft.com/k8s/aad-pod-identity/nmi:1.3"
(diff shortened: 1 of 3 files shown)


charts/aad-pod-identity/templates/infra.yaml
```diff
@@ -146,9 +147,10 @@ spec:
       {{- end }}
       containers:
       - name: mic
-        image: "mcr.microsoft.com/k8s/aad-pod-identity/mic:1.2"
+        image: "mcr.microsoft.com/k8s/aad-pod-identity/mic:1.3"
         imagePullPolicy: Always
         args:
+          - mic
```
kkmsft: What does name as parameter do ?

khenidak: we moved to alpine images with different docker file. This is just directly calling mic as container entry point (this is across mic/nmi on charts and deployment spec)

PR #152, opened 2019-02-01
Add back entrypoints to images

cpuguy83:


cpuguy83:
OH maybe should not have updated the helm chart until new images are available.

cpuguy83:
Should I just build and push a new set over existing tags?

/cc @khenidak

cpuguy83:
Maybe a 1.4.1 and 1.3.1 respectively?

kkmsft:
@cpuguy83 - I think we should revert the change, bring back only entry point change and then change the deployment files before the next release.

kkmsft:
May be we should think of releasing the deployment file or modify the release notes to state the deployment to use.. that way there is no confusion..

PR #155, opened 2019-02-06
Keep binary arg in helm charts

cpuguy83:


kkmsft:
shouldn't we update the deployment as well if that is the case ?
In any case, I have updated the release top point to the right deployment files:
https://github.com/Azure/aad-pod-identity/releases/tag/1.3.0-mic-1.4.0-nmi

cpuguy83:
I don't really follow.
What is the deployment in this case?

Issue #169, opened 2019-03-04
Unable to use pod identity inside init containers


yarinm:
I'm trying to use MSI in an init container and I get the error: 
`time="2019-03-04T06:36:21Z" level=error msg="missing podname for podip:11.6.0.45, match failed, ip:11.6.0.45 matching pods:&PodList{ListMeta:k8s_io_apimachinery_pkg_apis_meta_v1.ListMeta{SelfLink:/api/v1/pods,ResourceVersion:31310752,Continue:,},Items:[],}" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=11.6.0.45
`

After some testing it seems that problem is with NMI service and the way to queries the pod name [by pod ip](https://github.com/Azure/aad-pod-identity/blob/master/pkg/k8s/client.go#L80)

You can see that because it uses the field selector with the status.phase=Running it can't work in an init container since the pod status during that time is Pending.

I think this needs to change to status.phase==Running or status.phase==Pending. Note that you can use OR in field selector syntax. Only thing that seems to be working doing several not equals.

WDYT?

kkmsft:
cc @rite2nikhil

yarinm:
I've created a PR for this issue - https://github.com/Azure/aad-pod-identity/pull/170

kkmsft:
Thank you, @yarinm for the PR #191 . Closing with with the merge of that PR.

PR #170, opened 2019-03-05
Changed phase status filtering to work with init containers as well

yarinm:
Fix for https://github.com/Azure/aad-pod-identity/issues/169

Note that you can't use OR in selector interface but AND with NOT does work so i created an ignore list for statuses with don't want pod-identity to work on.


pkg/k8s/client.go
```diff
@@ -21,6 +22,9 @@ const (
 	getPodListSleepTimeMilliseconds = 300
 )
 
+// We only want to allow pod-identity with Pending or Running phase status
+var IGNORED_POD_PHASE_STATUSES = []string{"Succeeded", "Failed", "Unknown", "Completed", "CrashLoopBackOff"}
```
kkmsft: Can you please move this to different casing ? 
May be: var ignoredPodPhases = []string {...}

pkg/k8s/client.go
```diff
@@ -76,8 +80,9 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
 }
 
 func (c *KubeClient) getPodListWithTries(podip string, tries int) (*v1.PodList, error) {
+	phaseFilter := ",status.phase!=" + strings.Join(IGNORED_POD_PHASE_STATUSES, ",status.phase!=")
```
kkmsft: Is there a way we can do this without recomputing the filter every time ?

pkg/k8s/client.go
```diff
@@ -21,6 +22,9 @@ const (
 	getPodListSleepTimeMilliseconds = 300
 )
```
kkmsft: First of all - Thank you @yarinm for raising the issue and this PR.

pkg/k8s/client.go
yarinm: I’m no golang expert but I think there are two possible solutions here:
1. Write the filter string in a hard-coded way
2. Create a package wide variable For the filter and create a getOrCreateFilter method that will initialize the variable once and return its value

WDYT?

pkg/k8s/client.go
yarinm: @kkmsft I went over your comments and resolved the issues, let me know what you think.

yarinm:
@kkmsft ? any updates about this? can we merge it?

kkmsft:
@yarinm - Apologies for the delay. Initially I was hoping to get the e2e stabilized and then merge this. But that is going to take some time. Right now there is azure pipeline failure with this. I will fix that first and then get it merged this week. Thank you very much for this work.

kkmsft:
@yarinm - can you please rebase your branch to the latest ?

yarinm:
I'm closing this PR, merge https://github.com/Azure/aad-pod-identity/pull/191

PR #176, opened 2019-03-12
Fix vm type parsing

cpuguy83:
Before if the cluster config was vmss but the machine type was a vm, it
would still use a vmss.


cpuguy83:
rebased on top of #179, otherwise CI fails.

PR #179, opened 2019-03-15
Use go modules

cpuguy83:
This fixes issues with CI just using `go get` or having to fetch
external tools (such as `dep`).
The issue with ci using `go get` (without using go modules) is it
fetches the master/HEAD from the upstream repo which can easily (and
currently is) break.


cpuguy83:
ping @kkmsft 

CI is currently broken because `go get` is not compiling due to changed API's upstream.
With go modules `go build` and `go get` work the right way and get the deps you expect.

PR #203, opened 2019-05-01
Fix vmss identity deletion for ID in use

### LINKED ISSUE #198: VMSS: Removing identity from a node removes identity for the scaleset even if other nodes are using it
This was found during internal testing. As the function which determines InUse, does not check whether there are other nodes in the same VM Scale Set which uses the identity, currently in use identities can be removed. We need to fix the function checking in use to consider the scenario that an identity might still be in use by other nodes in the given VMScaleSet.

cpuguy83:
Before this change, an identity on a VMSS node may be deleted from the
VMSS even if it's used on another node in the VMSS.

This makes the check for identity in use vmss aware since identities on
a vmss are managed for the entire vmss, not just a single node in the
vmss.

Fixes #198

### DIFF
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -5,18 +5,15 @@ import (
 	"sync"
 	"time"
 
-	"github.com/Azure/aad-pod-identity/pkg/stats"
-	"github.com/Azure/aad-pod-identity/version"
-
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
-
 	"github.com/Azure/aad-pod-identity/pkg/cloudprovider"
 	"github.com/Azure/aad-pod-identity/pkg/crd"
 	"github.com/Azure/aad-pod-identity/pkg/pod"
+	"github.com/Azure/aad-pod-identity/pkg/stats"
+	"github.com/Azure/aad-pod-identity/version"
 	"github.com/golang/glog"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -118,7 +115,16 @@ func (c *Client) Start(exit <-chan struct{}) {
 
 func (c *Client) Sync(exit <-chan struct{}) {
 	glog.Info("Sync thread started.")
-	for event := range c.EventChannel {
+
+	var event aadpodid.EventType
+
+	for {
+		select {
+		case <-exit:
+			return
+		case event = <-c.EventChannel:
+		}
+
 		stats.Init()
 		// This is the only place where the AzureAssignedIdentity creation is initiated.
 		begin := time.Now()
@@ -151,6 +157,8 @@ func (c *Client) Sync(exit <-chan struct{}) {
 		}
 		stats.Put(stats.System, time.Since(systemTime))
 
+		nodeRefs := make(map[string]bool)
+
 		var newAssignedIDs []aadpodid.AzureAssignedIdentity
 		beginNewListTime := time.Now()
 		//For each pod, check what bindings are matching. For each binding create volatile azure assigned identity.
@@ -175,6 +183,7 @@ func (c *Client) Sync(exit <-chan struct{}) {
 				if allBinding.Spec.Selector == crdPodLabelVal {
 					glog.V(5).Infof("Found binding match for pod %s/%s with binding %s", pod.Name, pod.Namespace, allBinding.Name)
 					matchedBindings = append(matchedBindings, allBinding)
+					nodeRefs[pod.Spec.NodeName] = true
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/mic/mic.go
```diff
@@ -223,9 +239,36 @@ func (c *Client) Sync(exit <-chan struct{}) {
 		if deleteList != nil && len(*deleteList) > 0 {
 			beginDeletion := time.Now()
 			workDone = true
+
+			vmssNodes := make(map[string][]*corev1.Node)
```
kkmsft: Can we make it more generic ? or check if its only vmss then perform this ?

pkg/mic/mic.go
```diff
@@ -223,9 +239,36 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
+				if r.ResourceType == cloudprovider.VMSSResourceType {
+					id := makeVMSSID(r)
+					vmssNodes[id] = append(vmssNodes[id], n)
+				}
+			}
+
 			for _, delID := range *deleteList {
 				glog.V(5).Infof("Deletion of id: %s", delID.Name)
-				inUse := c.checkIfInUse(delID, newAssignedIDs)
+				inUse, err := c.checkIfInUse(delID, newAssignedIDs, vmssNodes)
```
kkmsft: Can we please add an e2e test to test the vmss id deletion ?

pkg/mic/mic.go
cpuguy83: So for each item in the list we need to check if the reference is against a vmss node and if any other node in the vmss is referencing it. This means we need to determine for each item if the node is a vmss.
Instead of doing this on demand I'm just gathering all the vmss nodes (which are being referenced) in one shot which prevents duplicate lookups.

In fact thinking about this more, this may not be enough... we may want to move adds before deletes so that we don't unnecessarily remove/re-add to a vmss node.

pkg/mic/mic.go
```diff
@@ -295,6 +318,47 @@ func (c *Client) Sync(exit <-chan struct{}) {
 	}
 }
 
+func (c *Client) getVMSSNodes(refs map[string]bool) (map[string][]*corev1.Node, error) {
+	ls := make(map[string][]*corev1.Node)
+	for ref := range refs {
+		n, err := c.NodeClient.Get(ref)
```
kkmsft: Can't we already identity whether vm is VMA or VMSS by the time we get the node client ?

pkg/mic/mic.go
```diff
@@ -223,9 +235,20 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
+				glog.Error(err)
+				continue
+			}
+
 			for _, delID := range *deleteList {
 				glog.V(5).Infof("Deletion of id: %s", delID.Name)
-				inUse := c.checkIfInUse(delID, newAssignedIDs)
+				inUse, err := c.checkIfInUse(delID, newAssignedIDs, vmssNodes)
+				if err != nil {
+					glog.Error(err)
```
kkmsft: We should delete this specific id from the del list, otherwise we will delete identities which are still in use, because we failed a cloud call.

pkg/mic/mic.go
cpuguy83: If we can I don't know where.

pkg/mic/mic.go
cpuguy83: Will update.

cpuguy83: Wait, this is where we are doing the deletion, there shouldn't be any accidental deletions.

pkg/mic/mic.go
```diff
@@ -175,6 +186,7 @@ func (c *Client) Sync(exit <-chan struct{}) {
 				if allBinding.Spec.Selector == crdPodLabelVal {
 					glog.V(5).Infof("Found binding match for pod %s/%s with binding %s", pod.Name, pod.Namespace, allBinding.Name)
 					matchedBindings = append(matchedBindings, allBinding)
+					nodeRefs[pod.Spec.NodeName] = true
```
kkmsft: What happens in a case where the binding got deleted. Wouldn't we miss adding the node on which the binding was applied. Is that going to be an issue ?

pkg/mic/mic.go
```diff
@@ -151,6 +160,8 @@ func (c *Client) Sync(exit <-chan struct{}) {
 		}
 		stats.Put(stats.System, time.Since(systemTime))
 
+		nodeRefs := make(map[string]bool)
```
kkmsft: can we please rename nodeRefs to something which represents the usage requirement-  as in we are trying to grab the nodes which potentially have identities which are still in use because a binding exists, so could we please  name this to reflect this usage.

pkg/mic/mic.go
```diff
@@ -470,10 +553,28 @@ func (c *Client) checkIfInUse(checkAssignedID aadpodid.AzureAssignedIdentity, ar
 		if checkID.Spec.Type != aadpodid.UserAssignedMSI {
 			continue
 		}
-		if checkID.Spec.ClientID == id.Spec.ClientID && checkAssignedID.Spec.NodeName == assignedID.Spec.NodeName &&
-			checkAssignedID.Spec.Pod != assignedID.Spec.Pod {
-			return true
+
+		if checkAssignedID.Spec.Pod == assignedID.Spec.Pod {
```
kkmsft: If the pod is the same, then shouldn't we check rest before continuing ? or did you want to do != here ?

pkg/mic/mic.go
cpuguy83: This happens before deletion so we are adding all nodes which have a binding at all.

pkg/mic/mic.go
cpuguy83: Do you have a suggestion?
`nodeRefs` fits the bill for me, especially since this is all nodes with a pod that has an identity.

pkg/mic/mic.go
cpuguy83: If the pod is the same then there's nothing to check, I think?
The old behavior was returning early if all those things matched.
This just says if it's the same pod then then continue on, which is what the old code did since it would reach the end of the loop rather than return early.

cpuguy83:
Updated as per our conversation.

PR #205, opened 2019-05-02
Update docs and err msg in tutorial script

aramase:
- Update broken docs url
- Few typos and update to tutorial docs


kkmsft:
Thank you @aramase

Issue #108, opened 2018-10-29
Pod retry - pending review comments #100


kkmsft:
Raising this issue to track the review comments from #100 
- Add the retry sleep interval as option.
- Add UT for the call.

kkmsft:
Closing via #208 merge.

PR #208, opened 2019-05-03
Add retry sleep interval as option

aramase:
Changes in reference to https://github.com/Azure/aad-pod-identity/issues/108

- Add retry sleep interval as an option
- Add UT for `GetPodName` and test retry logic


kkmsft:
Thank you @aramase

Issue #113, opened 2018-11-09
Worker node's IP tables aren't cleaned up after deleting CRDs


chewong:
Expected the aad metadata rules to be deleted from worker nodes' ip tables after deleting the CRDs (i.e. `kubectl delete -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml`).
Command executed in worker nodes: `sudo iptable -L -t nat`

aramase:
@kkmsft can you assign this issue to me?

PR #154, opened 2019-02-05
add uninstall notes

khenidak:
fixes https://github.com/Azure/aad-pod-identity/issues/113

PR #211, opened 2019-05-06
Handle iptables cleanup on uninstall

aramase:
resolves https://github.com/Azure/aad-pod-identity/issues/113

This will handle graceful shutdown of the pod, cleanup the ip table chain and rules that were inserted by NMI.

### DIFF
--- a/pkg/nmi/iptables/iptables.go
+++ b/pkg/nmi/iptables/iptables.go
@@ -115,7 +115,7 @@ func DeleteCustomChain() error {
 	if err != nil {
 		return err
 	}
-	if err := removeCustonChainReference(ipt, tablename, "PREROUTING"); err != nil {
+	if err := removeCustomChainReference(ipt, tablename, "PREROUTING"); err != nil {
 		return err
 	}
 	if err := removeCustomChain(ipt, tablename); err != nil {
@@ -124,8 +124,8 @@ func DeleteCustomChain() error {
 	return nil
 }
 
-// removeCustonChainReference - iptables -t "table" -D "chain" -j "customchainname"
-func removeCustonChainReference(ipt *iptables.IPTables, table, chain string) error {
+// removeCustomChainReference - iptables -t "table" -D "chain" -j "customchainname"
+func removeCustomChainReference(ipt *iptables.IPTables, table, chain string) error {
 	exists, err := ipt.Exists(table, chain, "-j", customchainname)
 	if err == nil && exists {
 		return ipt.Delete(table, chain, "-j", customchainname)
--- a/pkg/nmi/iptables/iptables.go
+++ b/pkg/nmi/iptables/iptables.go
@@ -108,24 +108,24 @@ func flushCreateCustomChainrules(ipt *iptables.IPTables, destIP, destPort, targe
 	return nil
 }
 
-// DeleteCustomChain deletes the custom chain aad-metadata reference from PREROUTING
-// chain and then deletes the chain aad-metadata
+// DeleteCustomChain removees the custom chain aad-metadata reference from PREROUTING
+// chain and then removes the chain aad-metadata from nat table
 func DeleteCustomChain() error {
 	ipt, err := iptables.New()
 	if err != nil {
 		return err
 	}
-	if err := deleteCustomChainInChain(ipt, tablename, "PREROUTING"); err != nil {
+	if err := removeCustonChainReference(ipt, tablename, "PREROUTING"); err != nil {
 		return err
 	}
-	if err := deleteCustomChain(ipt, tablename); err != nil {
+	if err := removeCustomChain(ipt, tablename); err != nil {
 		return err
 	}
 	return nil
 }
 
-// deleteCustomChainInChain - iptables -t "table" -D "chain" -j "customchainname"
-func deleteCustomChainInChain(ipt *iptables.IPTables, table, chain string) error {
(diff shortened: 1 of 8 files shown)


pkg/nmi/server/server.go
```diff
@@ -74,14 +77,27 @@ func (s *Server) Run() error {
...
-	log.Infof("node: %s %s", s.NodeName, s.HostIP)
-	for range time.Tick(time.Second * time.Duration(s.IPTableUpdateTimeIntervalInSeconds)) {
-		log.Infof("node(%s) hostip(%s) metadataaddress(%s:%s) nmiport(%s)", s.NodeName, s.HostIP, s.MetadataIP, s.MetadataPort, s.NMIPort)
-		if err := iptables.AddCustomChain(s.MetadataIP, s.MetadataPort, s.HostIP, s.NMIPort); err != nil {
-			log.Fatalf("%s", err)
-		}
-		if err := iptables.LogCustomChain(); err != nil {
-			log.Fatalf("%s", err)
+	signalChan := make(chan os.Signal)
+	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT)
```
kkmsft: Can you please add an e2e test which creates and deletes the deployment and checks whether the rules have been cleaned ?

pkg/nmi/server/server.go
```diff
@@ -346,3 +362,21 @@ func copyHeader(dst, src http.Header) {
...
+	log.Info("Received SIGTERM, shutting down")
+
+	exitCode := 0
+	// clean up iptables
+	if err := iptables.DeleteCustomChain(); err != nil {
+		log.Infof("Error cleaning up during shutdown: %v", err)
+		exitCode = 1
+	}
+
+	// wait for pod to delete
```
kkmsft: Which pod is this wait for ?

pkg/nmi/iptables/iptables.go
```diff
@@ -107,3 +107,41 @@ func flushCreateCustomChainrules(ipt *iptables.IPTables, destIP, destPort, targe
...
+		return err
+	}
+	if err := deleteCustomChain(ipt, tablename); err != nil {
+		return err
+	}
+	return nil
+}
+
+// deleteCustomChainInChain - iptables -t "table" -D "chain" -j "customchainname"
+func deleteCustomChainInChain(ipt *iptables.IPTables, table, chain string) error {
```
kkmsft: Could we please make the function name more clear - perhaps something like "removeCustomChainReference" or something that like that ?

kkmsft:
@aramase - Thank you for the PR. Have added some comments. Please take a look at them. 

An additional request is - could you also add the comments you made on manual deletion in #154 to documentation for cases where a manual delete might be required (for signals which are not actionable)

pkg/nmi/server/server.go
aramase: This is just wait for the graceful shutdown of the pod. This is not absolutely necessary as the termination grace period is 60s. So we will kill the container before the termination grace period.

test/e2e/aadpodidentity_test.go
```diff
@@ -446,6 +446,65 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 	// 	removeUserAssignedIdentityFromCluster(nodeList, clusterIdentity)
 	// })
 
+	It("should cleanup iptable rules after deleting aad-pod-identity", func() {
+		// create the helper ssh daemon for validating the iptable rules
+		cmd := exec.Command("kubectl", "apply", "-f", "template/busyboxds.yaml")
+		_, err := cmd.CombinedOutput()
+		Expect(err).NotTo(HaveOccurred())
+
+		time.Sleep(90 * time.Second)
```
aramase: Added 90s wait time because it's possible nmi pods are just being created. It takes 60s after the pod startup before the iptables entries are added since the ticker is configured with 60s update duration.  We might need to change this to create the ticker with instant first tick to prevent any race conditions.

kkmsft: Could we add a check to ensure that the rules have been indeed added instead of the sleep to make it more deterministic. Perhaps in a future PR, but could make our tests more deterministic ?

test/e2e/template/busyboxds.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  template:
+    metadata:
+      labels:
+        component: busybox
+    spec:
+      hostNetwork: true
+      containers:
+      - name: busybox
+        image: alpine:3.8
```
kkmsft: just wanted to confirm that its the newest version ?

test/e2e/aadpodidentity_test.go
aramase: That makes sense. I'll open an issue to track that change.

test/e2e/template/busyboxds.yaml
aramase: The latest version is `alpine:3.9.4`. I kept it at `3.8` because that is the version we are using to build other images.

test/e2e/template/busyboxds.yaml
```diff
@@ -0,0 +1,23 @@
...
+        stdin: true
+        securityContext:
+          privileged: true
```
kkmsft: do we need privileged for this ? Isn't hostNetwork: true enough to check the iptables existence ?

aramase: Yeah, without privileged the iptables table won't initialize.

PR #215, opened 2019-05-16
Add retries to role assignment

aramase:
@kkmsft I've set the retries to 10 times with wait interval of 6s. Read in another issue that the replication can take close to 2 mins in worst case scenario. This retry has mostly worked for me without any issues. We can increase the retries later if it still doesn't suffice.


test/e2e/setup.sh
```diff
@@ -22,15 +22,25 @@ IDENTITY_PRINCIPAL_ID=$(az identity show -g "$RESOURCE_GROUP" -n keyvault-identi
 az role assignment create --role 'Managed Identity Operator' --assignee "$AZURE_CLIENT_ID" --scope "/subscriptions/$SUBSCRIPTION_ID/resourcegroups/$RESOURCE_GROUP/providers/Microsoft.ManagedIdentity/userAssignedIdentities/keyvault-identity" || true
 az keyvault set-policy -n "$KEYVAULT_NAME" --secret-permissions get list --spn "$IDENTITY_CLIENT_ID" || true
 # The following command might need a couple of retries to succeed
-az role assignment create --role Reader --assignee "$IDENTITY_PRINCIPAL_ID" --scope "/subscriptions/$SUBSCRIPTION_ID/resourceGroups/$RESOURCE_GROUP/providers/Microsoft.KeyVault/vaults/$KEYVAULT_NAME" || true
+set +e
+for i in {0..10}; do
```
kkmsft: Please move the retry to a utility function to which you can pass the commands and use that in 3 places.

aramase: Done

kkmsft:
Thank you @aramase

PR #216, opened 2019-05-16
Log binary versions

### LINKED ISSUE #207: include binary versions in logs
Acceptance criteria: As a user, I would like to see binary versions of NMI and MIC in the logs.

CecileRobertMichon:
Acceptance criteria: As a user, I would like to see binary versions of NMI and MIC in the logs.

Fix a typo in the Makefile that caused the ARG for `NMI_VERSION` to be always empty. Add handling for `MIC_VERSION`.

These versions will show in the logs next time we build an image using `NMI_VERSION` and  `MIC_VERSION` args.

Fixes #207

### DIFF
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -15,6 +15,7 @@ var (
 	kubeconfig      string
 	cloudconfig     string
 	forceNamespaced bool
+	versionInfo     bool
 )
 
 func main() {
@@ -22,8 +23,12 @@ func main() {
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
+	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
 	flag.Parse()
-	log.Info("Starting mic process. Version: %v. Build date: %v", version.MICVersion, version.BuildDate)
+	if versionInfo {
+		version.PrintVersionAndExit()
+	}
+	glog.Info("Starting mic process. Version: %v. Build date: %v", version.MICVersion, version.BuildDate)
 	if cloudconfig == "" {
 		glog.Fatalf("Could not get the cloud config")
 	}
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/Azure/aad-pod-identity/pkg/mic"
+	"github.com/Azure/aad-pod-identity/version"
 	"github.com/golang/glog"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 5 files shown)


cmd/nmi/main.go
```diff
@@ -32,7 +33,7 @@ func main() {
 	if *debug {
 		log.SetLevel(log.DebugLevel)
 	}
-	log.Info("starting nmi process")
+	log.Info("Starting nmi process. Version: %v. Build date: %v", version.NMIVersion, version.BuildDate)
```
kkmsft: Can we please have the same in mic main.go which prints the version there ?

version/version.go
```diff
@@ -11,11 +11,14 @@ var BuildDate string
...
-// Version is the version of the binary
-var Version string
+// MICVersion is the version of the MIC component
+var MICVersion string
+
+// NMIVersion is the version of the NMI component
+var NMIVersion string
 
 // PrintVersionAndExit prints the version and exits
 func PrintVersionAndExit() {
```
kkmsft: Can this be used in case of -v command line option for both mmi and mic. Would be a good way to find out the comparable nmi/mic.

CecileRobertMichon: do we really want to exit the process after printing the version?

cmd/nmi/main.go
CecileRobertMichon: done

version/version.go
kkmsft: I was thinking if so if some one runs 

nmi -v
or 
mic -v

this is the only which needs to be called. In that case the exit can be left there as is - thoughts ?

Also wondering why the fmt was used here originally, does this assume that blog is not initialized or some other rule to be followed in version.go ?

cmd/mic/main.go
```diff
@@ -14,14 +15,20 @@ var (
...
 
 func main() {
 	defer glog.Flush()
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
+	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
 	flag.Parse()
+	if versionInfo {
+		version.PrintVersionAndExit()
```
CecileRobertMichon: is this what you meant @kkmsft ?

cmd/mic/main.go
```diff
@@ -14,14 +15,20 @@ var (
...
+	}
+	glog.Info("Starting mic process. Version: %v. Build date: %v", version.MICVersion, version.BuildDate)
```
kkmsft: Infof ?

cmd/nmi/main.go
```diff
@@ -29,10 +31,13 @@ var (
...
 func main() {
 	pflag.Parse()
+	if *versionInfo {
+		version.PrintVersionAndExit()
+	}
 	if *debug {
 		log.SetLevel(log.DebugLevel)
 	}
-	log.Info("starting nmi process")
+	log.Info("Starting nmi process. Version: %v. Build date: %v", version.NMIVersion, version.BuildDate)
```
kkmsft: Infof ?

PR #129, opened 2018-12-03
Update E2E to run tests based on generated images

pjbgf:
The former E2E tests were being executed against the publicly available docker images, instead of using the images generated locally and published into custom container registry. Making it easier for new contributors to support this project.

Environment Variables required:

- REGISTRY
- SUBSCRIPTION_ID
- RESOURCE_GROUP
- AZURE_CLIENT_ID

Once the environment variables are set, changes can be built and tested in a cluster with: 
`make build && make image && make push && make e2e`

PR #218, opened 2019-05-17
Update E2E to run tests based on generated images 

CecileRobertMichon:
Based on #129 from @pjbgf 

The former E2E tests were being executed against the publicly available docker images, instead of using the images generated locally and published into custom container registry. Making it easier for new contributors to support this project.

New ENV config variables:

REGISTRY 
NMI_VERSION
MIC_VERSION
IDENTITY_VALIDATOR_VERSION

Once the environment variables are set, changes can be built and tested in a cluster with:
make build && make image && make push && make e2e

### DIFF
--- a/test/common/k8s/deploy/deploy.go
+++ b/test/common/k8s/deploy/deploy.go
@@ -48,7 +48,7 @@ func CreateInfra(namespace, registry, nmiVersion, micVersion, templateOutputPath
 		return errors.Wrap(err, "Failed to parse deployment-rbac.yaml")
 	}
 
-	deployFilePath := path.Join(templateOutputPath, namespace+"-deployment-rbac.yaml")
+	deployFilePath := path.Join(templateOutputPath, namespace+"-deployment.yaml")
 	deployFile, err := os.Create(deployFilePath)
 	if err != nil {
 		return errors.Wrap(err, "Failed to create a deployment file from deployment-rbac.yaml")
--- a/test/common/k8s/deploy/deploy.go
+++ b/test/common/k8s/deploy/deploy.go
@@ -41,8 +41,47 @@ type Status struct {
 	AvailableReplicas int `json:"availableReplicas"`
 }
 
+// CreateInfra will deploy all the infrastructure components (nmi and mic) on a Kubernetes cluster
+func CreateInfra(namespace, registry, nmiVersion, micVersion, templateOutputPath string) error {
+	t, err := template.New("deployment-rbac.yaml").ParseFiles(path.Join("template", "deployment-rbac.yaml"))
+	if err != nil {
+		return errors.Wrap(err, "Failed to parse deployment-rbac.yaml")
+	}
+
+	deployFilePath := path.Join(templateOutputPath, namespace+"-deployment-rbac.yaml")
+	deployFile, err := os.Create(deployFilePath)
+	if err != nil {
+		return errors.Wrap(err, "Failed to create a deployment file from deployment-rbac.yaml")
+	}
+	defer deployFile.Close()
+
+	deployData := struct {
+		Namespace  string
+		Registry   string
+		NMIVersion string
+		MICVersion string
+	}{
+		namespace,
+		registry,
+		nmiVersion,
+		micVersion,
+	}
+	if err := t.Execute(deployFile, deployData); err != nil {
+		return errors.Wrap(err, "Failed to create a deployment file from deployment-rbac.yaml")
+	}
+
+	cmd := exec.Command("kubectl", "apply", "-f", deployFilePath)
+	util.PrintCommand(cmd)
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return errors.Wrap(err, "Failed to deploy Infrastructure to the Kubernetes cluster")
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


test/e2e/config/config.go
```diff
@@ -4,12 +4,16 @@ import "github.com/kelseyhightower/envconfig"
...
-	KeyvaultSecretVersion string `envconfig:"KEYVAULT_SECRET_VERSION"`
+	SubscriptionID           string `envconfig:"SUBSCRIPTION_ID"`
+	ResourceGroup            string `envconfig:"RESOURCE_GROUP"`
+	AzureClientID            string `envconfig:"AZURE_CLIENT_ID"`
+	KeyvaultName             string `envconfig:"KEYVAULT_NAME"`
+	KeyvaultSecretName       string `envconfig:"KEYVAULT_SECRET_NAME"`
+	KeyvaultSecretVersion    string `envconfig:"KEYVAULT_SECRET_VERSION"`
+	MICVersion               string `envconfig:"MIC_VERSION"`
+	NMIVersion               string `envconfig:"NMI_VERSION"`
+	Registry                 string `envconfig:"REGISTRY"`
```
kkmsft: Is this the variable - REGISTRY,  which will change where the image would get pulled from ?

test/e2e/config/config.go
```diff
@@ -4,12 +4,16 @@ import "github.com/kelseyhightower/envconfig"
...
-	KeyvaultSecretName    string `envconfig:"KEYVAULT_SECRET_NAME"`
-	KeyvaultSecretVersion string `envconfig:"KEYVAULT_SECRET_VERSION"`
+	SubscriptionID           string `envconfig:"SUBSCRIPTION_ID"`
+	ResourceGroup            string `envconfig:"RESOURCE_GROUP"`
+	AzureClientID            string `envconfig:"AZURE_CLIENT_ID"`
+	KeyvaultName             string `envconfig:"KEYVAULT_NAME"`
+	KeyvaultSecretName       string `envconfig:"KEYVAULT_SECRET_NAME"`
+	KeyvaultSecretVersion    string `envconfig:"KEYVAULT_SECRET_VERSION"`
+	MICVersion               string `envconfig:"MIC_VERSION"`
+	NMIVersion               string `envconfig:"NMI_VERSION"`
```
kkmsft: Can you please add some notes on the README on steps to run custom images.

test/e2e/config/config.go
CecileRobertMichon: correct. so for prod it would be `mcr.microsoft.com/k8s/aad-pod-identity/`

test/e2e/config/config.go
CecileRobertMichon: done

test/e2e/README.md
```diff
@@ -26,24 +26,41 @@ export KEYVAULT_SECRET_NAME='...'
...
 At the same time, the tests utilizes two user assigned identities, `keyvault-identity` (have read access to the keyvault that you create) and `cluster-identity` (have read access to the resource group level). You can create necessary Azure resources and roles with the bash script [`setup.sh`](./setup.sh) (Note that reader assignment in the script might need a few attempts to succeed).
 
 Finally, to start the e2e tests, execute the following commands:
+
 ```bash
 cd $GOPATH/src/github.com/Azure/aad-pod-identity
 
 # Ensure that the local project and the dependencies are in sync
-dep ensure
+make mod
```
CecileRobertMichon: updated since we switched to go modules

CecileRobertMichon:
@kkmsft this is ready for review

test/common/k8s/deploy/deploy.go
```diff
@@ -83,7 +126,7 @@ func Delete(name, templateOutputPath string) error {
 	util.PrintCommand(cmd)
 	_, err := cmd.CombinedOutput()
 	if err != nil {
-		return errors.Wrap(err, "Failed to delete AzureIdentityBinding from the Kubernetes cluster")
+		return errors.Wrapf(err, "Failed to delete %v from the Kubernetes cluster", name)
```
kkmsft: Isn't this create failure ?

CecileRobertMichon: no it's a delete operation (it's part of the `Delete()` func)

test/common/k8s/deploy/deploy.go
```diff
@@ -57,11 +96,15 @@ func CreateIdentityValidator(subscriptionID, resourceGroup, name, identityBindin
 
 	// Go template parameters to be translated in test/e2e/template/deployment.yaml
 	deployData := struct {
-		Name            string
-		IdentityBinding string
+		Name                     string
+		IdentityBinding          string
```
kkmsft: Do we want to take namespace here as well ?

kkmsft:
Thank you, @CecileRobertMichon

test/common/k8s/deploy/deploy.go
CecileRobertMichon: We could. Right now all the E2E test pods are created in the default namespace.

test/e2e/README.md
kkmsft: Thank you, @CecileRobertMichon

Issue #145, opened 2019-01-19
Identity assignment performance issues resulting in slow scale out


ankrause:
CloudProvider: Azure VMSS
Version: Latest from source

The below data is all based on VMSS, though I would guess that other scenarios like VMAS using user-assigned identities would be impacted as well.

I've been testing scale scenarios in a test cluster running VMSS built from AKS-Engine. Since the VMSS support commit hasn't been released yet, I've built and published images from current source to my own registry for testing purposes. From what I've seen, all operations are very fast except for CloudProvider.Put. Overall, it generally takes MIC a few seconds to create one identity binding.

```
I0119 21:29:17.404782       1 mic.go:314] Found 3 pods, 2 ids, 2 bindings
I0119 21:29:17.404820       1 stats.go:63] ** Stats collected **
I0119 21:29:17.404825       1 stats.go:59] Pod listing: 32.201µs
I0119 21:29:17.404831       1 stats.go:59] ID listing: 199.438823ms
I0119 21:29:17.404834       1 stats.go:59] Binding listing: 192.956583ms
I0119 21:29:17.404837       1 stats.go:59] Assigned ID listing: 201.481236ms
I0119 21:29:17.404860       1 stats.go:59] System: 593.916043ms
I0119 21:29:17.404863       1 stats.go:59] Cloud provider get: 31.137391ms
I0119 21:29:17.404866       1 stats.go:59] Cloud provider put: 38.190984112s
I0119 21:29:17.404869       1 stats.go:59] Assigned ID addition: 212.117301ms
I0119 21:29:17.404897       1 stats.go:59] Assigned ID deletion: 0s
I0119 21:29:17.404900       1 stats.go:59] Find assigned ids to create: 1.8µs
I0119 21:29:17.404904       1 stats.go:59] Find assigned ids to delete: 800ns
I0119 21:29:17.404907       1 stats.go:59] Total time to add assigned IDs: 38.434392305s
I0119 21:29:17.404910       1 stats.go:59] Total time to delete assigned IDs: 0s
I0119 21:29:17.404913       1 stats.go:59] Event recording: 20.7µs
I0119 21:29:17.404917       1 stats.go:59] Total: 39.028374148s
I0119 21:29:17.404920       1 stats.go:86] *********************
```

[...]

Issue #217, opened 2019-05-17
User Assigned Identity Not Removed from VM


droessmj:
Using AKS, Kubernetes version: 1.12.6

AAD Pod Identity Info:
- NMI: v1.4
- MIC: v1.3

##Symptoms
- Pods were stuck on a crashloopbackoff due to inability to authN to database using User Assigned Identity. MIC logs indicated that the correct binding was being applied to the pods in question, however the AzureIdentity resource referenced in MIC logs did not match the resources existing in Kubernetes. The AzureIdentity resource had been updated to consume a separate, newer User Assigned Identity resource in Azure. Logs indicated that the prior User Assigned Identity was attempting to be bound, but a 403 was being returned (as the prior identity resource was deleted from its resource group). 

##Error Message - Amended to remove identifying info
```
Event(v1.ObjectReference{Kind:"AzureIdentityBinding", Namespace:"default", Name:"MSI-BINDING-PLACEHOLDER", UID:"b4b93419-78ae-11e9-8dbb-e6d966016eaf", APIVersion:"aadpodidentity.k8s.io/v1", ResourceVersion:"5608855", FieldPath:""}): type: 'Warning' reason: 'binding apply error' Applying binding MSI-BINDING-PLACEHOLDER node aks-agentpool-ID-0 for pod APP-batch-7d96f9c87f-lgf97-default-msi-APP-mybatch1-d
resulted in error compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=403 -- Original Error: Code="LinkedAuthorizationFailed" Message="The client 'GUID' with object id 'GUID' has permission to perform action 'Microsoft.Compute/virtualMachines/write' on scope '/subscriptions/SUBSCRIPTION/virtualMachines/aks-agentpool-ID-0'; however, it does not have permission to perform action 'Microsoft.ManagedIdentity/userAssignedIdentities/assign/action' on the linked scope(s) '/subscriptions/SUBSCRIPTION8/resourcegroups/RG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/IDENTITY-Name'."
```

##Recorded steps performed to recreate and resolve
- New MSI was created and AKS Cluster SP was assigned Managed Identity Operator RBAC role for MSI
- New AzureIdentity and AzureIdentityBinding was created. Deployment was updated to consume new binding
- Pods failed to authN to an Azure SQL database. MIC logs showed references to old User Assigned Identity
[...]

Issue #181, opened 2019-04-01
Trying to use Pod Identity fails a number of times, before finally succeeding


woutervanvliet:
**Intro:** I'm trying to get Azure Pod Identity to work in our cluster, and am mostly succeeding (so far so good). For the time being, we have two keyvaults, two AzureIdentity's, two AzureIdentityBinding's and two Pods using each their keyvault. 

While testing, both pods are equal - only difference being their `aadpodidbinding` and an environment variable indicating what keyvault to use. At startup, the pod connects to the KeyVault, reads two values and prints them with `Console.WriteLine`. If the connection fails, the pod will crash and k8s will restart it.

**The problem:** One pod might startup being able to read from the keyvault immediately, while the other will crash and restart for - what seems to be - rather consistently 5 times before being able to get an access token.

When it fails, the following Exception is thrown:

```
Unhandled Exception: Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProviderException: Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/******************. Exception Message: Tried the following 3 methods to get an access token, but none of them worked.
Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/******************. Exception Message: Tried to get token using Managed Service Identity. Access token could not be acquired. MSI ResponseCode: Forbidden, Response: no AzureAssignedIdentity found for pod:default/kv-test-be

Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/******************. Exception Message: Tried to get token using Visual Studio. Access token could not be acquired. Environment variable LOCALAPPDATA not set.
Parameters: Connection String: [No connection string specified], Resource: https://vault.azure.net, Authority: https://login.windows.net/******************. Exception Message: Tried to get token using Azure CLI. Access token could not be acquired. No such file or directory

   at Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProvider.GetAuthResultAsyncImpl(String authority, String resource, String scope)
   at Microsoft.Azure.Services.AppAuthentication.AzureServiceTokenProvider.<get_KeyVaultTokenCallback>b__8_0(String authority, String resource, String scope)
   at Microsoft.Azure.KeyVault.KeyVaultCredential.PostAuthenticate(HttpResponseMessage response)
[...]

PR #219, opened 2019-05-17
Performance  Improvement - reduce identity assignment time

aramase:
resolves 
https://github.com/Azure/aad-pod-identity/issues/145
https://github.com/Azure/aad-pod-identity/issues/181
https://github.com/Azure/aad-pod-identity/issues/217

- Batch process id assignments to node on each sync cycle (pooling delete and add ids, issuing one CreateOrUpdate call for each sync cycle)
- Reduces identity assignment for 44 pods from 29m to ~3m

Add states to the assigned identities
1. Created - assigned identity will be in created state when the crd is created.
2. Assigned - once the identity has been successfully assigned to underlying node, the state of the assigned identity will be changed to Assigned.
3. Unassigned - the assigned identity will be in this state when the identity has been successfully removed from the underlying node.

**Note**
1. Assigned identities will be created and then the identity will be tried to be assigned to the node. If assignment fails, then the identity will be deleted. No change in this behavior compared to previous version.
2. Assigned identities will be deleted only after the identity has been successfully removed from the node. This is change in behavior from previous version to ensure we successfully remove the identities from the underlying node.

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -27,9 +27,12 @@ type Client struct {
 	Config     config.AzureConfig
 }
 
+// ClientInt client interface
 type ClientInt interface {
 	RemoveUserMSI(userAssignedMSIID string, node *corev1.Node) error
 	AssignUserMSI(userAssignedMSIID string, node *corev1.Node) error
+	UpdateUserMSI(addUserAssignedMSIIDs []string, removeUserAssignedMSIIDs []string, node *corev1.Node) error
+	GetUserMSIs(node *corev1.Node) ([]string, error)
 }
 
 // NewCloudProvider returns a azure cloud provider client
@@ -112,6 +115,59 @@ func withInspection() autorest.PrepareDecorator {
 	}
 }
 
+// GetUserMSIs will return a list of all identities on the node
+func (c *Client) GetUserMSIs(node *corev1.Node) ([]string, error) {
+	idH, _, err := c.getIdentityResource(node)
+	if err != nil {
+		glog.Errorf("GetUserMSIs: get identity resource failed with error %v", err)
+		return nil, err
+	}
+	info := idH.IdentityInfo()
+	if info == nil {
+		return nil, fmt.Errorf("identity info is nil")
+	}
+	idList := info.GetUserIdentityList()
+	return idList, nil
+}
+
+// UpdateUserMSI will batch process the removal and addition of ids
+func (c *Client) UpdateUserMSI(addUserAssignedMSIIDs []string, removeUserAssignedMSIIDs []string, node *corev1.Node) error {
+	idH, updateFunc, err := c.getIdentityResource(node)
+	if err != nil {
+		return err
+	}
+
+	info := idH.IdentityInfo()
+	if info == nil {
+		info = idH.ResetIdentity()
+	}
+
+	requiresUpdate := false
+	// remove msi ids from the list
+	for _, userAssignedMSIID := range removeUserAssignedMSIIDs {
+		requiresUpdate = true
+		if err := info.RemoveUserIdentity(userAssignedMSIID); err != nil {
+			return fmt.Errorf("could not remove identity from node %s: %v", node.Name, err)
+		}
+	}
+	// add new ids to the list
+	for _, userAssignedMSIID := range addUserAssignedMSIIDs {
+		addedToList := info.AppendUserIdentity(userAssignedMSIID)
+		if !addedToList {
+			glog.V(6).Infof("Identity %s already assigned to node %s. Skipping assignment.", userAssignedMSIID, node.Name)
+		}
+		requiresUpdate = requiresUpdate || addedToList
+	}
+	if requiresUpdate {
+		timeStarted := time.Now()
+		if err := updateFunc(); err != nil {
+			return err
+		}
+		glog.V(6).Infof("UpdateUserMSI of %s completed in %s", node.Name, time.Since(timeStarted))
+	}
+	return nil
+}
+
 //RemoveUserMSI - Use the underlying cloud api calls and remove the given user assigned MSI from the vm.
 func (c *Client) RemoveUserMSI(userAssignedMSIID string, node *corev1.Node) error {
 	idH, updateFunc, err := c.getIdentityResource(node)
@@ -137,6 +193,7 @@ func (c *Client) RemoveUserMSI(userAssignedMSIID string, node *corev1.Node) erro
 	return nil
 }
 
+// AssignUserMSI - Use the underlying cloud api call and add the given user assigned MSI to the vm
 func (c *Client) AssignUserMSI(userAssignedMSIID string, node *corev1.Node) error {
 	// Get the vm using the VmClient
 	// Update the assigned identity into the VM using the CreateOrUpdate
--- a/pkg/cloudprovider/identity.go
+++ b/pkg/cloudprovider/identity.go
@@ -36,7 +36,7 @@ func filterUserIdentity(idType *compute.ResourceIdentityType, idList *[]string, 
 	case compute.ResourceIdentityTypeUserAssigned,
 		compute.ResourceIdentityTypeSystemAssignedUserAssigned:
 	default:
-		return errNotFound
+		return nil
 	}
 
 	origLen := len(*idList)
--- a/pkg/cloudprovider/identity.go
+++ b/pkg/cloudprovider/identity.go
@@ -21,6 +21,7 @@ type IdentityHolder interface {
 type IdentityInfo interface {
 	AppendUserIdentity(id string) bool
 	RemoveUserIdentity(id string) error
+	GetUserIdentityList() []string
 }
 
 var (
@@ -78,10 +79,8 @@ func appendUserIdentity(idType *compute.ResourceIdentityType, idList *[]string, 
 	switch *idType {
 	case compute.ResourceIdentityTypeUserAssigned, compute.ResourceIdentityTypeSystemAssignedUserAssigned:
 		// check if this ID is already in the list
-		for _, id := range *idList {
-			if id == newID {
-				return false
-			}
+		if checkIfIDInList(*idList, newID) {
+			return false
 		}
 	case compute.ResourceIdentityTypeSystemAssigned:
 		*idType = compute.ResourceIdentityTypeSystemAssignedUserAssigned
@@ -91,4 +90,13 @@ func appendUserIdentity(idType *compute.ResourceIdentityType, idList *[]string, 
 
 	*idList = append(*idList, newID)
 	return true
+}
+
+func checkIfIDInList(idList []string, desiredID string) bool {
+	for _, id := range idList {
+		if id == desiredID {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/cloudprovider/vm.go
+++ b/pkg/cloudprovider/vm.go
@@ -115,3 +115,7 @@ func (i *vmIdentityInfo) AppendUserIdentity(id string) bool {
 	}
 	return appendUserIdentity(&i.info.Type, i.info.IdentityIds, id)
 }
+
+func (i *vmIdentityInfo) GetUserIdentityList() []string {
+	return *i.info.IdentityIds
+}
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -53,6 +53,12 @@ type ClientInt interface {
 	Sync(exit <-chan struct{})
 }
 
+type trackUserAssignedMSIIds struct {
+	addUserAssignedMSIIDs    []string
+	removeUserAssignedMSIIDs []string
+	node                     *corev1.Node
+}
+
 func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool) (*Client, error) {
 	glog.Infof("Starting to create the pod identity client. Version: %v. Build date: %v", version.MICVersion, version.BuildDate)
 
@@ -241,8 +247,14 @@ func (c *Client) Sync(exit <-chan struct{}) {
 
 		// Extract add list and delete list based on existing assigned ids in the system (currentAssignedIDs).
 		// and the ones we have arrived at in the volatile list (newAssignedIDs).
-		// TODO: Separate this into two methods.
-		addList, deleteList, err := c.splitAzureAssignedIDs(currentAssignedIDs, &newAssignedIDs)
+		// get list of new assigned ids to create
+		addList, err := c.getAzureAssignedIDsToCreate(currentAssignedIDs, &newAssignedIDs)
+		if err != nil {
+			glog.Error(err)
+			continue
+		}
+		// get list of assigned ids to delete
+		deleteList, err := c.getAzureAssignedIDsToDelete(currentAssignedIDs, &newAssignedIDs)
 		if err != nil {
 			glog.Error(err)
 			continue
@@ -250,8 +262,9 @@ func (c *Client) Sync(exit <-chan struct{}) {
 
 		glog.V(5).Infof("del: %v, add: %v", deleteList, addList)
 
+		nodeMap := make(map[string]trackUserAssignedMSIIds)
+
 		if deleteList != nil && len(*deleteList) > 0 {
-			beginDeletion := time.Now()
 			workDone = true
 
 			vmssGroups, err := getVMSSGroups(c.NodeClient, nodeRefs)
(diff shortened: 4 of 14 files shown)


kkmsft:
Thank you @aramase . A good e2e could be validating that #145 is resolved and the identities are assigned in a reasonable amount of time.

pkg/cloudprovider/identity.go
```diff
@@ -92,3 +91,12 @@ func appendUserIdentity(idType *compute.ResourceIdentityType, idList *[]string,
 	*idList = append(*idList, newID)
 	return true
 }
+
+func checkIfIDInList(idList *[]string, desiredID string) bool {
```
cpuguy83: No need for the slice pointer here.

pkg/cloudprovider/vm.go
```diff
@@ -107,3 +108,10 @@ func (i *vmIdentityInfo) AppendUserIdentity(id string) bool {
 	}
 	return appendUserIdentity(&i.info.Type, i.info.IdentityIds, id)
 }
+
+func (i *vmIdentityInfo) CheckUserIdentityExists(id string) error {
+	if i.info.IdentityIds == nil || !checkIfIDInList(i.info.IdentityIds, id) {
+		return fmt.Errorf("Identity %s not found", id)
```
cpuguy83: nit: go style typically has errors start with lower case

Also, can this interface return a bool instead?

pkg/mic/mic.go
```diff
@@ -401,53 +457,148 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+			isUserAssignedMSI := c.checkIfUserAssignedMSI(id)
+
+			if isNodeErrored(createID.Spec.NodeName) && isUserAssignedMSI {
+				node, err := c.NodeClient.Get(createID.Spec.NodeName)
+				if err != nil {
+					c.EventRecorder.Event(binding, corev1.EventTypeWarning, "get node error",
+						fmt.Sprintf("Lookup of node %s for pod %s resulted in error %v", createID.Spec.NodeName, createID.Name, err))
+					continue
+				}
+				err = c.CloudClient.CheckUserMSI(id.Spec.ResourceID, node)
```
cpuguy83: Do we need the new interface? Can remove just do the right thing?

aramase:
@kkmsft I was able to validate https://github.com/Azure/aad-pod-identity/issues/145 with my changes. Created close to 44 pods with 2 assigned identities. With version 1.3 it took about 29m to create all assigned identities and assign msi to node. With my changes, it takes only about a minute since it's batching all the updates with a single call. I've a lot of refactoring of my PR still though.

kkmsft:
Thank you @aramas. Good to know that it's moving in the right direction. After our initial discussion on the approach have not looked deeply into the changes. Please let me know when you want me to start looking at the changes. 

Also, any chance (I know I am being greedy here ;-)) you can convert that test to an e2e test so that we can track any regression in future.

aramase:
@kkmsft Sure, I'll add that as an e2e. Still working on testing a couple of things. I'll need to refactor the PR. I'll let you know when it's ready for review.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -112,6 +115,59 @@ func withInspection() autorest.PrepareDecorator {
 	}
 }
 
+// CheckUserMSI checks if the identity has been assigned to the node
+func (c *Client) CheckUserMSI(userAssignedMSIID string, node *corev1.Node) bool {
+	idH, _, err := c.getIdentityResource(node)
```
kkmsft: Arent' these calls costly ? in the sense that they have to ARM every time ?

pkg/mic/mic.go
```diff
@@ -261,24 +275,48 @@ func (c *Client) Sync(exit <-chan struct{}) {
 
 				glog.V(5).Infof("Initiating assigned id creation for pod - %s, binding - %s", createID.Spec.Pod, binding.Name)
 
-				err = c.createAssignedIdentityDeps(&createID, id, node)
+				err = c.createAssignedIdentity(&createID)
```
kkmsft: Is there a way we can perform this create when we start looking at the node map ? Since we don't track state of the assigned identities, we could have a large time gap between the time the assignment is created and then the node assignment was made. Is that a possibility ? If so, can we mitigate it some how without additional state ?

pkg/cloudprovider/cloudprovider.go
aramase: @kkmsft You're right! we can avoid the multiple calls to ARM by just getting the list once and then validating existence/non-existence of all assigned ids based on that list.

pkg/mic/mic.go
aramase: @kkmsft This mimics the current logic. Currently, we create the assigned identity, then make the request for node assignment. The timegap still remains the same with this change because we are batching all the assignments and making the single call. So the time for assignment still remains the same. There shouldn't be a long time gap here.

pkg/mic/mic_test.go
```diff
@@ -159,9 +163,11 @@ func (c *TestCloudClient) CompareMSI(nodeName string, userIDs []string) bool {
 
 func (c *TestCloudClient) PrintMSI() {
 	for key, val := range c.ListMSI() {
+		fmt.Printf("\nNode name: %s\n", key)
```
kkmsft: Is this debugging logs ?

aramase: Yeah, I'll clean that up.

aramase:
Tests passed - 

```
Ran 8 of 9 Specs in 1339.441 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 1 Skipped
--- PASS: TestAADPodIdentity (1339.44s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	1339.459s
```

@kkmsft

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+
+			if !nodeHasErrors || idExistsOnNode {
+				// the identity was successfully assigned to the node
+				c.EventRecorder.Event(binding, corev1.EventTypeNormal, "binding applied",
+					fmt.Sprintf("Binding %s applied on node %s for pod %s", binding.Name, createID.Spec.NodeName, createID.Name))
+
+				glog.V(5).Infof("Initiating assigned id creation for pod - %s, binding - %s", createID.Spec.Pod, binding.Name)
+
+				if isUserAssignedMSI {
+					err := c.createAssignedIdentity(&createID)
```
aramase: The assigned identity will only be created once the identity assignment on the node is successfully complete.

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+
+			idExistsOnNode := c.checkIfMSIExistsOnNode(id, delID.Spec.NodeName, nodeIdentityList)
+
+			if !nodeHasErrors || !idExistsOnNode {
+				// the identity was successfully removed from node
+				c.EventRecorder.Event(removedBinding, corev1.EventTypeNormal, "binding removed",
+					fmt.Sprintf("Binding %s removed from node %s for pod %s", removedBinding.Name, delID.Spec.NodeName, delID.Spec.Pod))
+
+				// remove assigned identity crd from cluster
+				if err := c.removeAssignedIdentity(&delID); err != nil {
```
aramase: Similarly, removing the assigned identity only when the identity has been successfully removed from the node

test/e2e/aadpodidentity_test.go
```diff
@@ -231,7 +231,10 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 			identityValidatorName := fmt.Sprintf("identity-validator-%d", i)
 
 			setUpIdentityAndDeployment(keyvaultIdentity, fmt.Sprintf("%d", i))
-			time.Sleep(5 * time.Second)
+
```
aramase: Removed the sleep, so the test can actually proceed once the assigned identity exists.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -112,6 +115,59 @@ func withInspection() autorest.PrepareDecorator {
...
+	info := idH.IdentityInfo()
+	if info == nil {
+		return nil, fmt.Errorf("identity info is nil")
+	}
+	idList := info.GetUserIdentityList()
+	return idList, nil
+}
+
+// AppendOrRemoveUserMSI will batch process the removal and addition of ids
+func (c *Client) AppendOrRemoveUserMSI(addUserAssignedMSIIDs []string, removeUserAssignedMSIIDs []string, node *corev1.Node) error {
```
kkmsft: Is it possible to split this into two separate functions, one which adds and another which removes and then we can add individually add UTs for them ?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -112,6 +115,59 @@ func withInspection() autorest.PrepareDecorator {
...
+			glog.V(6).Infof("Identity %s already assigned to node %s. Skipping assignment.", userAssignedMSIID, node.Name)
+		}
+		requiresUpdate = requiresUpdate || addedToList
+	}
+	if requiresUpdate {
+		timeStarted := time.Now()
+		if err := updateFunc(); err != nil {
+			return err
+		}
+		glog.V(6).Infof("CreateOrUpdate of %s completed in %s", node.Name, time.Since(timeStarted))
```
kkmsft: This log seems out of place ?

pkg/mic/mic.go
```diff
@@ -211,46 +217,54 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
 		glog.V(5).Infof("del: %v, add: %v", deleteList, addList)
 
+		nodeMap := make(map[string]trackUserAssignedMSIIds)
+
 		if deleteList != nil && len(*deleteList) > 0 {
-			beginDeletion := time.Now()
 			workDone = true
 			for _, delID := range *deleteList {
 				glog.V(5).Infof("Deletion of id: %s", delID.Name)
+				// The inUse here checks if there are pods which are using the MSI in the newAssignedIDs.
```
kkmsft: nit: MSI ==> Identity

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+	err := c.CRDClient.RemoveAssignedIdentity(assignedID)
+	if err != nil {
+		glog.Error(err)
+	}
+	return nil
+}
+
+func (c *Client) appendToRemoveListForNode(nodeMap map[string]trackUserAssignedMSIIds, resourceID string, node *corev1.Node) {
+	if trackList, ok := nodeMap[node.Name]; ok {
+		if trackList.removeUserAssignedMSIIDs != nil {
```
kkmsft: Is this check necessary ?

pkg/cloudprovider/cloudprovider.go
kkmsft: Or may be naming can be UpdateUserMSI - AppendOrRemove feels like CreateOrUpdate which is doing a different thing.

pkg/mic/mic.go
```diff
@@ -259,26 +273,39 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
+
+			// generate unique list so we don't make multiple calls to assign/remove same id
+			n.addUserAssignedMSIIDs = c.getUniqueIDs(n.addUserAssignedMSIIDs)
+			n.removeUserAssignedMSIIDs = c.getUniqueIDs(n.removeUserAssignedMSIIDs)
+
+			err = c.CloudClient.AppendOrRemoveUserMSI(n.addUserAssignedMSIIDs, n.removeUserAssignedMSIIDs, n.node)
+			if err != nil {
+				// check which all identity assignment failed
+				// remove those assigned identities
+				// TODO check with identity team if CreateOrUpdate results in error, is it all or some failed
```
kkmsft: Can we do this before the PR gets merged.

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+	}
+}
 
-	if id.Spec.Type == aadpodid.UserAssignedMSI {
+func (c *Client) handleNodeErrors(nodesWithError []string, addList *[]aadpodid.AzureAssignedIdentity, deleteList *[]aadpodid.AzureAssignedIdentity, newAssignedIDs []aadpodid.AzureAssignedIdentity) error {
 
-		err = c.CloudClient.AssignUserMSI(id.Spec.ResourceID, node)
+	nodeIdentityList := make(map[string][]string)
+	for _, n := range nodesWithError {
+		// get the list of msi's currently on node
```
kkmsft: msi => user assigned identities.

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
 		if err != nil {
-			glog.Error(err)
-			newErr := c.CRDClient.RemoveAssignedIdentity(assignedID)
-			if newErr != nil {
-				glog.Errorf("Error when removing assigned identity in create error path err: %v", newErr)
+			glog.Errorf("Lookup of node %s resulted in error %v", n, err)
+			continue
+		}
+		idList, err := c.getUserMSIListForNode(node)
+		if err != nil {
```
kkmsft: What is the plan here. If we just log and continue then idList can be having random values..?

pkg/mic/mic.go
```diff
@@ -401,57 +441,166 @@ func (c *Client) makeAssignedIDs(azID *aadpodid.AzureIdentity, azBinding *aadpod
...
+		nodeMap[node.Name] = trackUserAssignedMSIIds{addUserAssignedMSIIDs: []string{resourceID}, node: node}
+	}
+}
 
-	if id.Spec.Type == aadpodid.UserAssignedMSI {
+func (c *Client) handleNodeErrors(nodesWithError []string, addList *[]aadpodid.AzureAssignedIdentity, deleteList *[]aadpodid.AzureAssignedIdentity, newAssignedIDs []aadpodid.AzureAssignedIdentity) error {
 
-		err = c.CloudClient.AssignUserMSI(id.Spec.ResourceID, node)
+	nodeIdentityList := make(map[string][]string)
+	for _, n := range nodesWithError {
```
kkmsft: What happens if we crash or get restarted at this time ?

test/e2e/aadpodidentity_test.go
```diff
@@ -626,6 +626,8 @@ func setUpIdentityAndDeployment(azureIdentityName, suffix, replicas string) {
 	ok, err = daemonset.WaitOnReady(nmiDaemonSet)
 	Expect(err).NotTo(HaveOccurred())
 	Expect(ok).To(Equal(true))
+
+	time.Sleep(30 * time.Second)
```
aramase: we can remove this sleep in the states PR because there we can start waiting for n assigned identities with desired state.

pkg/mic/mic.go
aramase: getUserMSIListForNode returns nil on error. So we'll fail all of them.

pkg/mic/mic.go
aramase: Sent out an email.

pkg/cloudprovider/cloudprovider.go
aramase: Fixed

pkg/cloudprovider/cloudprovider.go
aramase: Renamed the method

aramase:
Screen Shot 2019-06-25 at 12 40 56 PM

pkg/mic/mic.go
```diff
@@ -336,6 +254,169 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
+				// one associated with this azure identity.
+				glog.V(5).Infof("%s identity not found when using %s binding", binding.Spec.AzureIdentity, binding.Name)
+			}
+		}
+	}
+	return newAssignedIDs, nodeRefs, nil
+}
+
+// getListOfIdsToDelete will go over the delete list to determine if the id is required to be deleted
+// only user assigned identity not in use are added to the remove list for the node
```
kkmsft: Nit: two sentences.

aramase:
Ran 10 of 10 Specs in 2353.337 seconds
--- PASS: TestAADPodIdentity (2353.34s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	2353.353s

Issue #204, opened 2019-05-01
Pod Identity should not use azure.json


khenidak:
Please switch Pod Identity to use its own configuration. Azure Pod Identity should depend on its own configuration/identity to run `mic`. 

ref: https://github.com/kubernetes/kubernetes/issues/77309

@kkmsft

kkmsft:
@feiyushi had already started working on this.

kkmsft:
Released with https://github.com/Azure/aad-pod-identity/releases/tag/v1.5

PR #221, opened 2019-05-20
Remove dependency on azure.json

feiyushi:
As discussed in #204, aad-pod-identity is to use its own configuration to run `mic`. In this implementation, it uses a k8s secret to store the information that `mic` needs traditionally from the azure.json.


kkmsft:
Thank you @feiyushi for this contribution.  In general /lgtm from my side. 

A few notes:
- The understanding is that the docs for usage will come in as another  PR. 
- For e2e, we will have to have another setting (via an env variable or so) to use the new deployment model and also have the setup steps in setup.sh(in the e2e directory) for setting up the secrets corresponding to this.

kkmsft:
Thank you, @feiyushi. Merging this PR. We can do the doc and E2E test in the next PR.

PR #222, opened 2019-05-20
docs: change acs-engine references to aks-engine

CecileRobertMichon:
acs-engine is deprecated and replaced by aks-engine.

Also changed "e2e" to upper case "E2E".


kkmsft:
Thank you, @CecileRobertMichon

Issue #182, opened 2019-04-02
NMI calling API Server for Pod Info for every call to IMDS: Add Caching


abhisheknsit:
NMI is hitting api-server to fetch pod info for each call to it for fetching UserAssignedIdentity token. 
This could lead to 2 problems:

1. Bad tenants on a cluster, could DoS the api-server easily using this.
2. Incase connectivity to api-server breaks, then datapath operations of tenant pods will get affected.

One of the Solution could be to have a caching layer in NMI, to serve pod info for the identities.

khenidak:
Hi. ADAL client libraries by definition cache the token. They will not get-new-token, unless the application is specifically asking for one.  We really never designed this for multi tenant environments. However your concern is some how  valid for non-multi-tenants. For that i would strongly advice to adjust qps/qps-burst on your api server. As for caching, I will add it as a feature request for vNext.

abhisheknsit:
Hi @khenidak , We wouldn't want to trust the services deployed on the cluster, to follow best practices. Some of the services are using UserAssignedIdentites before the SDK support was there, so they are still having rest api implementations.

I didn't get the multi-tenant part. For me a tenant is a single application, and using aad-pod-identity I am able to stop one tenant from accessing other tenants Identities.

More than the use case of users revoking identities, I am worried about the use case of Service Disruption if the connectivity to api server is broken for some reason.

I absolutely, understand caching is tricky and we might need to move some part of IdentityBinding informers from MIC to NMI, but I think to keep a system with large number of identities stable it is a requirement.

Since the number of apicalls to the apiserver can get larger, VMSS can support a large number of identities, each identity can have access to multiple resources and each Service can have multiple pods.

kkmsft:
ListPodIds call made from NMI eventually calls the crdClient ListPodIds. With #232 we use go client cache listing to get the list of pods and assigned identities, thus reducing the load on API Server. Closing this bug with the merge of #232

kkmsft:
Another call in NMI is still pending to be fixed  - getPodListWithTries has direct client calls to get the pods. This call also needs to go to cache instead of direct API Server to satisfy the requirements in this bug report.

kkmsft:
Closing this since all listers have been converted to caching now. @abhisheknsit - Thank you for raising the issue. please re-open if there is more to be fixed related to the issue at hand.

PR #223, opened 2019-05-22
[WIP] Cache listers for aadpodidentity CRDs

kkmsft:
- Utilize the client cache for listing of CRDs
- Avoids going to API Server for every call.
- Resolve the part of #182 related to the API Server calls.


kkmsft:
Closing. Instead raising #232

PR #225, opened 2019-05-23
fix deployment template for e2e

aramase:
This is needed because `mic:1.3` and `nmi:1.4` required the `- mic` and `- nmi` args. I'll remove these during next release.


kkmsft:
Thank you, @aramase

Issue #224, opened 2019-05-23
MSI failed to assign identity to the underlying vm on Azure China AKS


northtyphoon:
We tried to set up MSI on AKS running in Azure China. The controller pod failed to assign the identity on the underlying VM. The following is the error.

E0522 23:19:58.096035       1 vm.go:66] compute.VirtualMachinesClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="SubscriptionNotFound" Message="The subscription 'xxx' could not be found."

It looks the reason is the resource manage endpoint is hard coded to Azure Public Cloud.

https://github.com/Azure/aad-pod-identity/blob/57f375703d8c6a33e53223eefe22af20ab0e3c9c/pkg/cloudprovider/vm.go#L27

kkmsft:
Thank you @northtyphoon for raising the issue. Its indeed a limitation.  We could look at command line parameters and arrive at the conclusion about the endpoint to contact. Would you be interested in submitting a PR for fixing this ?

kkmsft:
Other options and to look into the azure.json as current or include that into the set of secret reference instead of azure.json (add one more param here in #221 changes)

northtyphoon:
I will try to do a PR after we finished the deployment on China AKS. Right now we workaround it by manually assign the identity to the underlying vm. After that, the MSI works fine.

PR #226, opened 2019-05-23
Determine resource mgr endpoint based on cloud name

aramase:
resolves https://github.com/Azure/aad-pod-identity/issues/224


kkmsft:
Thank you, @aramase

PR #231, opened 2019-06-03
docs: tighten and organize README.md

mboersma:
This tries to make the project README more concise and approachable for someone who isn't very familiar with AAD Pod Identity (like me).

* Adds badges for go docs and the Go Report Card (A+)
* Sorts the sections so **Getting Started** is first
* Puts the **Contributing** section in a separate CONTRIBUTING.md document as recommended by GitHub's [community guidelines](https://github.com/Azure/aad-pod-identity/community)
* Moves the **Design**, **Tutorial**, **Debugging**, and **Contributing** sections into links in a **What To Do Next?** section
* Adds an example of the `az identity create` output
* Lots of markdown linter cleanup
* Moves markdown target links to the bottom for DRY and readability
* Tries to streamline as much of the writing as possible
* Moves **Uninstall Notes** to the end of **Getting Started**

This is pretty hard to read as a diff, so [this view](https://github.com/Azure/aad-pod-identity/blob/21ff1d596ee5b13e6eb45347e245672f248fac05/README.md) might be easier.

cc: @craiglpeters

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,18 +26,18 @@ You will need a Kubernetes cluster running on Azure, either managed by [AKS] or 
 
 ### 1. Create the Deployment
 
-AAD Pod Identity consists of the Managed Identity Controller (MIC) and Node Managed Identity (NMI) components in a Kubernetes deployment. For more information, see [Components].
+AAD Pod Identity consists of the Managed Identity Controller (MIC) deployment, the Node Managed Identity (NMI) daemon set, and several standard and custom resources. For more information, see [Components].
 
 Run this command to create the `aad-pod-identity` deployment on an RBAC-enabled cluster:
 
 ```shell
-kubectl create -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml
+kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml
 ```
 
 Or run this command to deploy to a non-RBAC cluster:
 
 ```shell
-kubectl create -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment.yaml
+kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment.yaml
 ```
 
 ### 2. Create an Azure Identity
@@ -86,7 +86,7 @@ Replace the placeholders with your user identity values. Set `type: 0` for user-
 Finally, save your changes to the file, then create the `AzureIdentity` resource in your cluster:
 
 ```shell
-kubectl create -f aadpodidentity.yaml
+kubectl apply -f aadpodidentity.yaml
 ```
 
 ### 4. (Optional) Match Pods in the Namespace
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,221 +1,274 @@
...
 
-#remove the custom chain
-iptables -t nat -X aad-metadata
-```
+Specifically, when a pod is scheduled, the MIC assigns an identity to the underlying VM during the creation phase. When the pod is deleted, it removes the assigned identity from the VM. The MIC takes similar actions when identities or bindings are created or deleted.
+
+### Node Managed Identity
+
+The authorization request to fetch a Service Principal Token from an MSI endpoint is sent to a standard Instance Metadata endpoint which is redirected to the NMI pod. The redirection is accomplished by adding rules to redirect POD CIDR traffic with metadata endpoint IP on port 80 to the NMI endpoint. The NMI server identifies the pod based on the remote address of the request and then queries Kubernetes (through MIC) for a matching Azure identity. NMI then makes an Azure Active Directory Authentication Library ([ADAL]) request to get the token for the client id and returns it as a response. If the request had client id as part of the query, it is validated against the admin-configured client id.
```
mboersma: I made some minor changes here, but I think this is still a confusing section that could use rewriting. Maybe in the next pass...

README.md
```diff
@@ -1,221 +1,274 @@
...
-* [Components](#components)
-* [Getting Started](#getting-started)
-* [Demonstration](#demonstration)
-* [Tutorial](#tutorial)
-* [Debugging](#debugging)
-* [Contributing](#contributing)
+AAD Pod Identity enables Kubernetes applications to access cloud resources securely with [Azure Active Directory] (AAD).
 
-## Design
+Using Kubernetes primitives, administrators configure identities and bindings to match pods. Then without any code modifications, your containerized applications can leverage anything in the cloud that depends on AAD as an identity provider.
```
kkmsft: nit: anything => any resource ?

README.md
```diff
@@ -1,221 +1,274 @@
...
+AAD Pod Identity consists of the Managed Identity Controller (MIC) and Node Managed Identity (NMI) components in a Kubernetes deployment. For more information, see [Components].
 
-```bash
-curl http://127.0.0.1:2579/host/token/?resource=https://vault.azure.net -H "podname: nginx-flex-kv-int" -H "podns: default"
-```
+Run this command to create the `aad-pod-identity` deployment on an RBAC-enabled cluster:
 
-## Getting Started
+```shell
+kubectl create -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml
```
ritazh: Should we use `apply` instead?

README.md
```diff
@@ -1,221 +1,274 @@
...
+```
 
-### Prerequisites 
+Or run this command to deploy to a non-RBAC cluster:
 
-A running k8s cluster on Azure using AKS or AKS Engine 
+```shell
+kubectl create -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment.yaml
```
ritazh: same

README.md
```diff
@@ -1,221 +1,274 @@
...
+  ResourceID: /subscriptions/<subid>/resourcegroups/<resourcegroup>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>
+  ClientID: <clientId>
 ```
 
-```
+Replace the placeholders with your user identity values. Set `type: 0` for user-assigned MSI or `type: 1` for Service Principal.
+
+Finally, save your changes to the file, then create the `AzureIdentity` resource in your cluster:
+
+```shell
```
ritazh: same

README.md
```diff
@@ -1,221 +1,274 @@
...
-### Understanding Namespaced identities
-The system will match `pod` to `identity` across namespaces by default. This behavior can be modified to match to pods within the namespace that holds `AzureIdentity` by
+### 4. (Optional) Match Pods in the Namespace
 
-1. On Azure Identity basis, by adding `aadpodidentity.k8s.io/Behavior: namespaced` annotation (You have to add the annotation on each `AzureIdentity` you want to apply this behavior on).
-2. Default namespaced behavior on all identities by adding `--forceNamespaced=true` argument on the command line  or declare `FORCENAMESPACED=true` environment variable (for both `nmi` and `mic`).
+By default, AAD Pod Identity matches pods to identities across namespaces. To match only pods in the namespace containing `AzureIdentity`, use one of these techniques:
 
+* Attach a `aadpodidentity.k8s.io/Behavior: namespaced` [annotation] to each `AzureIdentity` resource.
+* Add the `--forceNamespaced=true` command line argument or set the `FORCENAMESPACED=true` environment variable when starting both the MIC and NMI components.
```
ritazh: would be helpful to see an example yaml of this

README.md
```diff
@@ -1,221 +1,274 @@
...
-
+  AzureIdentity: <a-idname>
+  Selector: <label value to match>
 ```
+
+Replace the placeholders with your values. Ensure that the `AzureIdentity` name matches the one in `aadpodidentity.yaml`.
+
+Finally, save your changes to the file, then create the `AzureIdentityBinding` resource in your cluster:
+
+```shell
```
ritazh: apply?

README.md
```diff
@@ -1,221 +1,274 @@
...
 ```
-### Starting the demo pod
 
-Update the `deploy/demo/deployment.yaml` arguments with your subscription, clientID and resource group.
+### Start the Demo Pod
 
-```
+Update the `deploy/demo/deployment.yaml` arguments with your subscription, clientID and resource group, then create the demo deployment:
+
+```shell
```
ritazh: apply

README.md
mboersma: I can't decide... `kubectl apply` is more flexible and probably preferable in general, but this is a "Getting Started" section where arguably `kubectl create` is preferable because we are explicitly creating new resources and might want an error rather than overwriting them.

It's easy to change this if you have a strong preference, although I think we would also want to change the preceding wording to "create or update" in those cases.

README.md
mboersma: Good idea, I'll add something.

README.md
mboersma: I changed it to `apply` after more consideration.

README.md
```diff
@@ -1,221 +1,308 @@
...
 
-## Contributing
+```bash
+curl http://127.0.0.1:2579/host/token/?resource=https://vault.azure.net -H "podname: nginx-flex-kv-int" -H "podns: default"
+```
 
-This project welcomes contributions and suggestions.  Most contributions require you to agree to a
-Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us
-the rights to use your contribution. For details, visit https://cla.microsoft.com.
+### Uninstall Notes
```
kkmsft: Uninstall notes seems out of place. I am thinking it should be moved up right after the getting started or right after the demo instead of after components.

mboersma: Good idea. I moved it.

kkmsft:
Thank you @mboersma

PR #232, opened 2019-06-03
List CRDs via cache and avoid extra work on pod update

kkmsft:
Two main items:
        	- Utilize the client cache for listing of CRDs
		- Avoids going to API Server for every list call.
		- Resolve the part of #182 related to the API Server calls.
	- Avoid calling sync loop at every pod update. Only call where there is a change in scheduled node name.

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -20,9 +20,9 @@ import (
 type Client struct {
 	rest                *rest.RESTClient
 	BindingListWatch    *cache.ListWatch
-	BindingWatcher      cache.SharedInformer
+	BindingInformer     cache.SharedInformer
 	IDListWatch         *cache.ListWatch
-	IdWatcher           cache.SharedInformer
+	IDInformer          cache.SharedInformer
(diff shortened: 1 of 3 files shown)


pkg/crd/crd.go
```diff
@@ -18,9 +18,12 @@ import (
...
 type Client struct {
-	rest           *rest.RESTClient
-	BindingWatcher cache.SharedInformer
-	IdWatcher      cache.SharedInformer
+	rest                *rest.RESTClient
+	BindingListWatch    *cache.ListWatch
+	BindingWatcher      cache.SharedInformer
+	IDListWatch         *cache.ListWatch
+	IdWatcher           cache.SharedInformer
+	AssignedIDListWatch *cache.ListWatch
```
khenidak: The naming of `..Watch` gives the impression that they are all the same type which the are not. Can we use a different naming for informers vs watchers?

kkmsft: Thank you @khenidak.

Issue #213, opened 2019-05-15
Make e2e test deterministic


aramase:
https://github.com/Azure/aad-pod-identity/pull/211#discussion_r284441515

Issue #214, opened 2019-05-15
Removing identity from vm when system assigned identity is present fails


kkmsft:
When a vm is already having system assigned identity, then removing the user identity from it fails. This causes one of the e2e tests to fail with the following error:

compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="ResourceIdentity identityIds should not be provided if the type is \"SystemAssigned\"." Target="resourceIdentity.identityIds"

kkmsft:
Fixed with #259

PR #239, opened 2019-06-11
Make e2e tests deterministic

aramase:
**Reason for Change**:

- Make e2e deterministic. Remove sleep and wait until requirements are met.
- Fix tests
- Resolve panics in e2e
- Skip known failing test

**Issue Fixed**:

resolves https://github.com/Azure/aad-pod-identity/issues/213

**Notes for Reviewers**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,8 @@ else
 endif
 
 GO_BUILD_OPTIONS := --tags "netgo osusergo"  -ldflags "-s -X $(NMI_VERSION_VAR)=$(NMI_VERSION) -X $(MIC_VERSION_VAR)=$(MIC_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE) -extldflags '-static'"
-E2E_TEST_OPTIONS := -count=1 -v -timeout 24h -ginkgo.failFast
+# TODO (@aramase) remove the skip test once https://github.com/Azure/aad-pod-identity/issues/214 is resolved
+E2E_TEST_OPTIONS := -count=1 -v -timeout 24h -ginkgo.failFast -ginkgo.skip="should not alter the system assigned identity after creating and deleting pod identity"
 
 # useful for other docker repos
 REGISTRY_NAME ?= upstreamk8sci
--- a/test/common/k8s/azureassignedidentity/azureassignedidentity.go
+++ b/test/common/k8s/azureassignedidentity/azureassignedidentity.go
@@ -63,7 +63,9 @@ func WaitOnLengthMatched(target int) (bool, error) {
 				list, err := GetAll()
 				if err != nil {
 					errorChannel <- err
+					return
 				}
+				// len of nil slices is 0, so shouldn't panic here
 				if len(list.Items) == target {
 					successChannel <- true
 					return
--- a/test/common/k8s/daemonset/daemonset.go
+++ b/test/common/k8s/daemonset/daemonset.go
@@ -85,6 +85,7 @@ func WaitOnReady(name string) (bool, error) {
 				match, err := isAvailableReplicasMatchDesired(name)
 				if err != nil {
 					errorChannel <- err
+					return
 				}
 				if match {
 					successChannel <- true
--- a/test/common/k8s/daemonset/daemonset.go
+++ b/test/common/k8s/daemonset/daemonset.go
@@ -1,0 +1,107 @@
+package daemonset
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"time"
+
+	"github.com/Azure/aad-pod-identity/test/common/util"
+	"github.com/pkg/errors"
+)
+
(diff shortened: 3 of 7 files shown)


test/common/k8s/daemonset/daemonset.go
```diff
@@ -0,0 +1,107 @@
...
+	fmt.Printf("# Poll to check if %s daemonset is ready...\n", name)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				errorChannel <- errors.Errorf("Timeout exceeded (%s) while waiting for daemonset (%s) to be available", duration.String(), name)
+			default:
+				match, err := isAvailableReplicasMatchDesired(name)
+				if err != nil {
+					errorChannel <- err
```
kkmsft: As we discussed yesterday, we need to have a way to ensure that in case of error this go func concludes its execution right away.

test/e2e/aadpodidentity_test.go
```diff
@@ -322,9 +329,9 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 	// })
 
 	It("should not alter the user assigned identity on VM after AAD pod identity is created and deleted", func() {
-		azureIdentityResource := "/subscriptions/%s/resourceGroups/aad-pod-identity-e2e/providers/Microsoft.ManagedIdentity/userAssignedIdentities/%s"
-		clusterIdentityResource := fmt.Sprintf(azureIdentityResource, cfg.SubscriptionID, clusterIdentity)
-		keyvaultIdentityResource := fmt.Sprintf(azureIdentityResource, cfg.SubscriptionID, keyvaultIdentity)
+		azureIdentityResource := "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.ManagedIdentity/userAssignedIdentities/%s"
+		clusterIdentityResource := fmt.Sprintf(azureIdentityResource, cfg.SubscriptionID, cfg.ResourceGroup, clusterIdentity)
```
kkmsft: Thank you @aramase for including this.

test/e2e/aadpodidentity_test.go
```diff
@@ -699,3 +719,54 @@ func removeSystemAssignedIdentityOnCluster(nodeList *node.List) {
...
+	fmt.Println("# Tight-poll to check if iptable rules for aad-metadata exist on all nodes...")
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				errorChannel <- errors.Errorf("Timeout exceeded (%s) while waiting for iptable rules to exist", duration.String())
+			default:
+				list, err := pod.GetAllNameByPrefix(prefix)
+				if err != nil {
+					errorChannel <- err
```
kkmsft: Similar pattern here as well.

test/common/k8s/daemonset/daemonset.go
aramase: @kkmsft was able to simulate and try out few scenarios. We can either break/return after we encounter error and write to the channel. That way we prevent rest of the block from being executed and in the mean time the receiver gets the error and returns. I've used return everywhere just to stay in sync with what we already have for the success channel.

test/common/k8s/azureassignedidentity/azureassignedidentity.go
```diff
@@ -63,7 +63,9 @@ func WaitOnLengthMatched(target int) (bool, error) {
 				list, err := GetAll()
 				if err != nil {
 					errorChannel <- err
+					return
 				}
+				// len of nil slices is 0, so shouldn't panic here
```
aramase: The panic here was happening because when there is an error in `GetAll()` we return nil for list. With the return on error that should be resolved.

Makefile
```diff
@@ -35,7 +35,8 @@ else
 endif
 
 GO_BUILD_OPTIONS := --tags "netgo osusergo"  -ldflags "-s -X $(NMI_VERSION_VAR)=$(NMI_VERSION) -X $(MIC_VERSION_VAR)=$(MIC_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE) -extldflags '-static'"
-E2E_TEST_OPTIONS := -count=1 -v -timeout 24h -ginkgo.failFast
+# TODO (@aramase) remove the skip test once https://github.com/Azure/aad-pod-identity/issues/214 is resolved
+E2E_TEST_OPTIONS := -count=1 -v -timeout 24h -ginkgo.failFast -ginkgo.skip="should not alter the system assigned identity after creating and deleting pod identity"
```
aramase: Adding this so others don't hit a known failure while running e2e. I'll remove it once we fix issue #214

kkmsft: Thank you @aramase

test/e2e/aadpodidentity_test.go
kkmsft: return ?

aramase: Sorry, missed that! Added.

kkmsft:
Thank you, @aramase

Issue #237, opened 2019-06-10
Use a selector for AzureIdentityBinding


ritazh:
**Describe the request**
Use a selector for AzureIdentityBinding instead of match pod on a static label

**Explain why AAD Pod Identity needs it**
Use selector semantics to provide more flexibility

**Describe the solution you'd like**
As a user, I would like to use standard label selector semantics in AzureIdentityBinding to get matching pods.

For example:
```yaml
apiVersion: "aadpodidentity.k8s.io/v1"
kind: AzureIdentityBinding
metadata:
  name: test-azure-id-binding
spec: 
  AzureIdentity: "test-azure-identity"
  selector: 
    aadpodidbinding: select_it
```

**Describe alternatives you've considered**

**Additional context**

cc @kkmsft @aramase

cnadolny:
I'll be helping with this issue :)

Are there any E2E tests in place that need to be changed or run to verify the fix? I saw that crd_test and mic_test reference the selector, are there any others?

The following fixes need to be made:
- [ ] Edit existing YAML files
- [ ] Edit Go files, from a cursory glance looks like types.go needs fixing
- [ ] Update tests
- [ ] Update existing documentation

Let me know if there's anything I missed. I'll update as I make progress or come across anything else.

aramase:
@cnadolny The e2e tests are in here - https://github.com/Azure/aad-pod-identity/tree/master/test/e2e along with instructions on how to run them. 

Let us know if you have any questions. Looking forward to your PR :)

kkmsft:
@cnadolny - Thank you for looking into this. One important item with regard to this is backward capability - older installations should work with the new code.

cnadolny:
> @cnadolny - Thank you for looking into this. One important item with regard to this is backward capability - older installations should work with the new code.

To clarify with backwards compatibility - already existing pods bound to identities should continue to function, but should AzureIdentityBinding be expected to accept both static labels AND label selectors to work with older versions YAML files? 

Also, for the label selector - is it only accepting keys labeled with aadpodidbinding? Or are other label keys or multiple keys allowed to be accepted as well?

tkporter:
Hi! I have a use case that this change would make possible-- I see that https://github.com/Azure/aad-pod-identity/pull/244 was closed, but was hoping someone could give any suggestions for a workaround :)

[...]

PR #244, opened 2019-06-13
Implement label selector for AzureIdentityBinding

cnadolny:
**Reason for Change**:
Use a label selector for AzureIdentityBinding for greater flexibility

**Issue Fixed**:
#237

**Notes for Reviewers**:
I know that backwards compatibility is a necessity - but I'm a little confused on how to accomplish that. In order to implement the label selector, the Selector type for AzureIdentityBinding needs to be changed to type map[string]string from type string. But both YAML formats below should be accepted in order to support older versions, any pointers or suggestions?

Also - the changes would require the mic image to be rebuilt and pushed. Is that something I have to do from my end?

```
kind: AzureIdentityBinding
metadata:
  name: test-azure-id-binding
spec: 
  AzureIdentity: "test-azure-identity"
  selector: 
    aadpodidbinding: select_it
```
```
kind: AzureIdentityBinding
metadata:
  name: test-azure-id-binding
spec: 
  AzureIdentity: "test-azure-identity"
  selector: select_it
```


khenidak:
Thank you for this. I am concerned without our current approach. Maybe we can:
1- Create an internal api version, not visible to users
2- Create a new versioned api (v2?) that has the selector
3- MIC will need to watch for both versions, and convert to the internal version, which will then operate on.

Eventually once conversion and defaulting webhooks https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/ becomes stable we can remove the conversion code and update the CRD spec to use them.

pkg/apis/aadpodidentity/v1/types.go
```diff
@@ -155,7 +155,7 @@ const (
 type AzureIdentityBindingSpec struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	AzureIdentity     string `json:"azureidentity"`
-	Selector          string `json:"selector"`
+	Selector          map[string]string `json:"selector"`
```
ritazh: Can we use the actual `LabelSelector` type: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L1076

so that we can get the label from the selector and we can validate user's input:
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/validation/validation.go#L30

cnadolny:
> Thank you for this. I am concerned without our current approach. Maybe we can:
> 1- Create an internal api version, not visible to users
> 2- Create a new versioned api (v2?) that has the selector
> 3- MIC will need to watch for both versions, and convert to the internal version, which will then operate on.
> 
> Eventually once conversion and defaulting webhooks https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/ becomes stable we can remove the conversion code and update the CRD spec to use them.

Hi Khal - thanks for the suggestion!

Few questions:
- Where would this internal API live? Same area path as V1/V2? 
- I'm not sure if it's possible to declare two versions of a CRD API right now:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions

"For the first example, the assumption is all versions share the same schema with no conversion between them"

As I understand it, only the webhook conversion allows versions with differing schemas, which is only available as an alpha feature as you mentioned.

I tried to make a CRD definition with the two versions, like so:
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: azureidentitybindings.aadpodidentity.k8s.io
spec:
  group: aadpodidentity.k8s.io
  versions:
  - name: v2
    served: true
    storage: true
  - name: v1
    served: true
    storage: false
  version: v2
  names:
    kind: AzureIdentityBinding
    plural: azureidentitybindings
  scope: Namespaced
  ```

But whenever I tried to create an object of type v1:
  ```
apiVersion: aadpodidentity.k8s.io/v1
kind: AzureIdentityBinding
metadata:
  name: keyvault-identity-binding
spec:
  AzureIdentity: keyvault-identity
  Selector: keyvault-identity
```

Kubernetes ignores the apiVersion and creates it as v2. So it doesn't look like we can support both schemas and have MIC watch for both, unless I'm missing or misunderstanding something?

What I've committed so far:
- Created V2 API. Identical to V1, apart from new LabelSelector to AzureIdentityBindingSpec.
- Added watchers for both V1/V2, updated logic in client to store both in a map.
- Added code to convert V1 to V2 and return all as one list of V2 objects in ListBindings(). 
  - I can convert both of these to some internal API format easily but I wasn't sure where to put the internal API. 
  - I have not been able to test this code because of the issue above with not being able to declare two versions with differing schemas.
- Changed selector from type map[string]string to LabelSelector. Implemented logic to MatchLabels, still have to do MatchExpressions.

cnadolny:
My questions from the comment above were clarified after a call with Kal :) I've updated my changes after our discussion. This is still a WIP PR.

Since there are a lot of changes, I was thinking that it might be helpful to split these changes out into two PRs. One to create the internal API, and a second PR that creates the V2 API and implements the label selector. 

Current changes:
- Created internal API
  - Changed all aadpodid references in project from v1 to internal
  - Moved utils from V1/V2 to internal API since it is shared code 
- Created V2 API
  - Added new field called `LabelSelector` of type `Kubernetes LabelSelector` in AzureIdentityBinding struct
  - Original field `Selector` of type `string` is still contained in the struct to ensure backwards compatibility
- Created a conversion function to convert V1 API to Internal API in `utils.go`. This will convert the original static label into a Kubernetes LabelSelector.
  - The best place for conversions to take place would be in Sync(). This is the only place in the code where AzureIdentityBindings, AzureIdentities, and AzureAssignedIdentities are grabbed from the k8 cache.ListWatch.
  - Currently, I feel that only a conversion from V1 -> Internal API is needed, as the LabelSelector/Selector are only used to find matching pods to bindings, and create azure assigned identities for those pods.
- Updated logic in `Sync` function in `mic.go` to check pods to matching bindings based on LabelSelector rather than static label.
- Moved event types from `types.go` to `utils.go`
- Changed IdentityType from `int` to `string`

Left to do:
- Tests
  - Add/Modify Unit Tests
  - Add/Modify E2E Tests
- Documentation updates
- Rebase and get new changes

pkg/apis/aadpodidentity/utils.go
```diff
@@ -0,0 +1,72 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+//TODO: is this the right place for the conversion functions?
+//TODO: V2 and internal are currently the same API. 
+
+func ConvertV1ToInternal(identityBinding *[]AzureIdentityBinding, assignedIdentity *[]AzureAssignedIdentity, identity *[]AzureIdentity) (resIdentityBinding *[]AzureIdentityBinding, resAssignedIdentity *[]AzureAssignedIdentity, resIdentity *[]AzureIdentity) {
```
khenidak: Why do we have one func for all types?

khenidak: Also needs a unit tests please.

pkg/pod/pod.go
```diff
@@ -5,14 +5,15 @@ import (
 
 	"github.com/Azure/aad-pod-identity/pkg/stats"
 
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
+	// "k8s.io/apimachinery/pkg/labels"
+	// "k8s.io/apimachinery/pkg/selection"
```
khenidak: please remove commented import

pkg/apis/aadpodidentity/utils.go
cnadolny: Moved internal api + utils file to pkg/internal
Moved conversion code into own file and created new directory pkg/conversion, added unit tests, split up into functions by type like suggested

cnadolny:
Update on this. 

Refactoring the project to use the internal API has been a little trickier than expected. I've run into a few issues:
- The project needs to be able to listen for resources of multiple versions. RESTClientFor in `crd.go` takes in a version. Changing to UnversionedRestClientFor solves that, but I haven't been able to figure out how to successfully add multiple schemes (one schema for each type, since each scheme also takes in a version as well) to later be able to listen for resources of each version. 
- `cache.NewSharedInformer` takes in a versioned type as well. I think the project would need to be refactored to take in multiple informers, but have not tested.
- Currently the project retrieves CRD resources via *cache.ListWatch in `crd.go`. When you list the resources, for example in ListBindings(), the ListWatch instance returns back a single runtime.Object of type AzureIdentityBindingList, which is versioned as well, either v1 or v2. So I think this would also be a problem as it would cast all bindings into one version or the other.

I want to suggest splitting this out into two PRs. 
- One that creates a new V2 API, changes the project to reference and operate on that API. The API would include both a LabelSelector and Selector field for AzureIdentityBinding, which would satisfy the backwards compatibility requirement of the problem.
- A second PR/issue to finish refactoring the project to use the internal API.

lodejard:
At the risk of increasing the scope or size of this PR - as long as the objects are going through a v1/v2 change anyway... Would it be possible to change the AzureIdentitySpec.IdentityType from an int to a string? Could that be done in a way where the code treats existing values "0" and "1" as the equivalent "userAssigned" or "servicePrincipal" string values?

cnadolny:
Hi @kkmsft and @aramase ,

I have some time to work on this issue again. Is this issue still relevant? Want to check before jumping back in.

aramase:
@cnadolny The versioning changes PR has already been merged, and once we do add the v2 api it'll be in a separate PR. So can we close this PR?

cnadolny:
> @cnadolny The versioning changes PR has already been merged, and once we do add the v2 api it'll be in a separate PR. So can we close this PR?

Yes, this PR is outdated. Closed!

aramase:
Thank you @cnadolny!

PR #246, opened 2019-06-14
docs: add standalone Code of Conduct doc where GH can see it

mboersma:
**Reason for Change**:
Adds the Microsoft Open Source Code of Conduct as a standalone document in the project root, as expected by GitHub [community standards](https://github.com/Azure/aad-pod-identity/community).

**Issue Fixed**:
None, but see also Azure/aks-engine#510

**Notes for Reviewers**:


kkmsft:
@mboersma - is there a way to add a link to this and place that link as separate file as requirement ?

mboersma:
@kkmsft the trick is that Microsoft requires us to have the final paragraph of the README link to our online Code of Conduct, but GitHub wants to see a standalone Markdown document. 

So the requirements don't sync up perfectly, but I'll add a link in the README to the local version as well.

mboersma:
Or did you mean having CODE_OF_CONDUCT.md just contain a link to the online version? I thought of that and it would ensure we don't repeat ourselves. But I'm guessing that GitHub's intent is to have the project be self-documenting so that a tarball download contained this info.

The simplest solution seemed to be to allow ourselves to repeat the document, but let me know what you'd prefer.

kkmsft:
@mboersma - First Thank you :-) for looking at this and providing the context. I am leaning towards the first choice you mention above -  CODE_OF_CONDUCT.md just contain a link to the online version. Would prefer that it's a consistent pointer and not repeat ourselves.

kkmsft:
Thank you, @mboersma

PR #251, opened 2019-06-18
Fix incorrect resource endpoint with service principal

zhiweiv:
**Reason for Change**:
Resource endpoint for service principal was hard coded to https://management.azure.com/, so we are unable to use token to auth azure services such as Azure SQL, we should use the endpoint from request to generate the token.

**Issue Fixed**:
n/a


zhiweiv:
@kkmsft

kkmsft:
Thank you, @zhiweiv for the PR.

PR #252, opened 2019-06-18
fix: handle new OAuthConfig error consistently

mboersma:
**Reason for Change**:
The error from `adal.NewOAuthConfig` was ignored, so the subsequent error handling block could never be entered since it was comparing the previous `err` to `nil` again. I also made the handling consistent with the rest of the function.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -62,11 +62,11 @@ func NewCloudProvider(configFile string) (c *Client, e error) {
 		glog.Errorf("Get cloud env error: %+v", err)
 		return nil, err
 	}
-	oauthConfig, _ := adal.NewOAuthConfig(azureEnv.ActiveDirectoryEndpoint, azureConfig.TenantID)
+	oauthConfig, err := adal.NewOAuthConfig(azureEnv.ActiveDirectoryEndpoint, azureConfig.TenantID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -62,11 +62,11 @@ func NewCloudProvider(configFile string) (c *Client, e error) {
 		glog.Errorf("Get cloud env error: %+v", err)
 		return nil, err
 	}
-	oauthConfig, _ := adal.NewOAuthConfig(azureEnv.ActiveDirectoryEndpoint, azureConfig.TenantID)
+	oauthConfig, err := adal.NewOAuthConfig(azureEnv.ActiveDirectoryEndpoint, azureConfig.TenantID)
 	if err != nil {
-		return nil, fmt.Errorf("creating the OAuth config: %v", err)
+		glog.Errorf("creating the OAuth config: %+v", err)
```
aramase: nit: Can we also change the error message to be `creating the OAuth config error` following the same pattern as rest of the error messages in that func?

mboersma: For sure–I updated it to `Create OAuth config error: %+v` to be totally parallel with the other local error messages.

aramase: Thanks @mboersma

kkmsft:
Thank you, @mboersma

PR #255, opened 2019-06-21
[WIP] Add states to assigned identity

aramase:
**Reason for Change**:

Add states to assigned identity
- Created: When the assigned crd is created
- Assigned: The identity is successfully assigned to node
- Unassigned: When the identity is successfully removed from node

**Issue Fixed**:

resolves https://github.com/Azure/aad-pod-identity/issues/256

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/identity.go
+++ b/pkg/cloudprovider/identity.go
@@ -35,7 +35,7 @@ func filterUserIdentity(idType *compute.ResourceIdentityType, idList *[]string, 
 	case compute.ResourceIdentityTypeUserAssigned,
 		compute.ResourceIdentityTypeSystemAssignedUserAssigned:
 	default:
-		return errNotFound
+		return nil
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/identity.go
```diff
@@ -35,7 +35,7 @@ func filterUserIdentity(idType *compute.ResourceIdentityType, idList *[]string,
 	case compute.ResourceIdentityTypeUserAssigned,
 		compute.ResourceIdentityTypeSystemAssignedUserAssigned:
 	default:
-		return errNotFound
+		return nil
```
kkmsft: We should check for TypeNone and return nil from there, but keep the default as is.

aramase:
Closing this PR in favor of https://github.com/Azure/aad-pod-identity/pull/219

Issue #260, opened 2019-06-22
Problem in PR #240


zhiweiv:
**Describe the bug**
After #240, getPodListWithTries in client.go returns all pods. Seems FieldSelector is overridden in 
https://github.com/kubernetes/client-go/blob/master/tools/cache/listwatch.go#L73

**Steps To Reproduce**
Build from master, try get token from nmi.

**Expected behavior**
getPodListWithTries returns the only pod with specified pod ip.

**AAD Pod Identity version**
master

kkmsft:
Thank you, @zhiweiv.  
We just noticed in that in one of our end to end systems and was going to raise the bug. Thank you for the pointer to the client-go code related to this. The strange thing we noticed was that this was not reproducible in some of the systems and it gives the correct pod info. Any idea why this would occur on some system and not others ?

kkmsft:
Also, we are using this: k8s.io/client-go v0.0.0-20180424211516-33f2870a2b83

zhiweiv:
That is strange, since ListWatch always uses the fieldSelector from NewListWatchFromClient to override the FieldSelector of ListOptions, I think it can be reproduced 100%.

```go
podListWatch := cache.NewListWatchFromClient(
		clientset.CoreV1().RESTClient(),
		"pods",
		v1.NamespaceAll,
		fields.Everything())
```

```go
func NewListWatchFromClient(c Getter, resource string, namespace string, fieldSelector fields.Selector) *ListWatch {
	optionsModifier := func(options *metav1.ListOptions) {
		options.FieldSelector = fieldSelector.String()
	}
	return NewFilteredListWatchFromClient(c, resource, namespace, optionsModifier)
}
```

```go
func NewFilteredListWatchFromClient(c Getter, resource string, namespace string, optionsModifier func(options *metav1.ListOptions)) *ListWatch {
	listFunc := func(options metav1.ListOptions) (runtime.Object, error) {
		optionsModifier(&options)
		return c.Get().
			Namespace(namespace).
			Resource(resource).
			VersionedParams(&options, metav1.ParameterCodec).
			Do().
			Get()
	}
	watchFunc := func(options metav1.ListOptions) (watch.Interface, error) {
		options.Watch = true
		optionsModifier(&options)
		return c.Get().
			Namespace(namespace).
			Resource(resource).
			VersionedParams(&options, metav1.ParameterCodec).
			Watch()
	}
[...]

PR #262, opened 2019-06-24
Change to filtered list watch to fix fieldselector

aramase:
**Reason for Change**:

`NewListWatchFromClient` doesn't respect the fieldselector options provided to filter pods. Instead we'll create `NewFilteredListWatchFromClient` with empty options modifier. In this case ListWatch will use the options we provide during list.

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/260

**Notes for Reviewers**:

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -68,12 +68,12 @@ func NewKubeClient() (Client, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


aramase:
Screen Shot 2019-06-24 at 2 01 42 PM

kkmsft:
Thank you, @aramase.

PR #265, opened 2019-06-25
Adds support for MIC to authenticate with azure using system assigned/user assigned MSI

kkmsft:
This PR adds the capability for MIC to look at azure.json or environment variables
to determine whether the system assigned or user assigned MSI has to be used for accessing
azure resources. The MIC requests for token based on MSI. Also contains changes in NMI to determine if the request is originating from an MIC replicaset. If so, NMI directly generates the tokens instead of looking up the azure assigned identity for the pod-binding match.

**Reason for Change**:

Adds ability for MIC to authenticate using system assigned/user assigned MSI.

**Issue Fixed**:

Resolves the item in #261.

**Notes for Reviewers**:
TODO: run the e2e on system assigned identity cluster.

### DIFF
--- a/docs/readmes/README.msi.md
+++ b/docs/readmes/README.msi.md
@@ -18,9 +18,14 @@ and also operations on the user assigned identity.
 
 After the cluster is created to perform the following steps to obtain the principal id:
 for VMAs:
-```az vm identity show -g <resource group> -n <vm name> -o yaml```
+
+```bash
+az vm identity show -g <resource group> -n <vm name> -o yaml
+```
 for VMSS:
-```az vmss identity show -g <resource group>  -n <vmss scalset name> -o yaml```
+```bash
+az vmss identity show -g <resource group>  -n <vmss scalset name> -o yaml
+```
 
 The type in the output of the above command will identify the system assigned or user assigned MSI. Please record the corresponding
 principal id.
--- a/docs/readmes/README.msi.md
+++ b/docs/readmes/README.msi.md
@@ -1,0 +1,41 @@
+## Introduction
+
+The MIC component in aad-pod-identity needs to authenticate with the cloud to assign and remove user assign identities onto
+virtual machines (VMAs) or virtual machine scale sets(VMSS). This authentication is performed using either the cluster credentials
+obtained from azure.json in AKS/aks-engine clusters or credentials given via environment variables.
+
+MIC can authenticate using the following options:
+1. Service principal
+2. System assigned MSI
+3. User assigned MSI
+
+The rest of the README describes the prerequisite role assignments to be performed and how to configure MIC to use system assigned/user assigned MSI.
+
+## Pre-requisites - role assignments
+MIC is responsible for performing operations such as assigning user assigned identity to the underlying vm or vmss which makes up the
+nodes in the Kubernetes cluster. The system/user assigned MSI needs to have role assignments authorizing such operations on the vms/vmss
+and also operations on the user assigned identity.
+
+After the cluster is created to perform the following steps to obtain the principal id:
+for VMAs:
+`az vm identity show -g <resource group> -n <vm name> -o yaml`
+for VMSS:
+`az vmss identity show -g <resource group>  -n <vmss scalset name> -o yaml`
+
+The type in the output of the above command will identify the system assigned or user assigned MSI. Please record the corresponding
+principal id.
+
+For creating a role assignment to authorize assignment/removal of user assigned identities on VMS/VMSS, run the following command:
+`az role assignment create --role "Contributor" --assignee <principal id from az vm/vmss identity command>  --scope /subscriptions/<sub id>/resourcegroups/<resource group name>`
+
+Now to ensure that the operations are allowed on individual identity, perform the following for every identity in use:
+`az role assignment create --role "Managed Identity Operator" --assignee <principal id from az vm/vmss identity command>  --scope /subscriptions/<subscription id>/resourcegroups/<resource group name>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<identity name>`
+
+
+## Authentication method
+In case the azure.json is used, the following keys indicates whether the cluster is configured with system assigned or user assigned identity:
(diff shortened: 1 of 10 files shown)


pkg/k8s/client.go
```diff
@@ -81,23 +81,28 @@ func NewKubeClient() (Client, error) {
...
-		return "", "", err
+		return "", "", "", err
 	}
 	numMatching := len(podList.Items)
 	if numMatching == 1 {
-		return podList.Items[0].Namespace, podList.Items[0].Name, nil
+		// TODO: Filter out the deployment owner references.
+		deployment := ""
+		if podList.Items[0].OwnerReferences[0].Kind == "ReplicaSet" {
+			deployment = podList.Items[0].OwnerReferences[0].Name
```
aramase: This is the replicaset name, so maybe we should change deployment to rsName to reflect the value we are getting

pkg/nmi/server/server.go
```diff
@@ -226,20 +227,53 @@ func (s *Server) msiHandler(logger *log.Entry, w http.ResponseWriter, r *http.Re
...
+		var token *adal.Token
+		// UserAssignedIdentity clientID is empty, so we are going to use system assigned MSI
+		if rqClientID == "" {
+			token, err = auth.GetServicePrincipalTokenFromMSI(rqResource)
+		} else { // User assigned identity usage.
+			token, err = auth.GetServicePrincipalTokenFromMSIWithUserAssignedID(rqClientID, rqResource)
+		}
+		if err != nil {
+			logger.Errorf("failed to get service principal token for pod:%s/%s, %+v", podns, podname, err)
+			http.Error(w, err.Error(), http.StatusForbidden)
```
aramase: Do we want to return the status code returned by metadata ep instead of 403 since we are really just proxying the request here? I think metadata ep will return a `400` or `404`. 403 is not part of retry status codes.

kkmsft: Not sure yet, how to extract this info from ada. For time being have put a TODO to handle in a future PR.

docs/readmes/README.msi.md
```diff
@@ -0,0 +1,50 @@
+## Introduction
+
+The MIC component in aad-pod-identity needs to authenticate with the cloud to assign and remove user assign identities onto
```
aramase: s/user assign/user assigned

docs/readmes/README.msi.md
```diff
@@ -0,0 +1,50 @@
...
+3. User assigned MSI
+
+The rest of the README describes the prerequisite role assignments to be performed for using MSI and how to configure MIC to use system assigned/user assigned MSI.
+
+## Pre-requisites - role assignments
+MIC is responsible for performing operations such as assigning user assigned identity to the underlying vm or vmss which makes up the
+nodes in the Kubernetes cluster. The system/user assigned MSI needs to have role assignments authorizing such operations on the vms/vmss
+and also operations on the user assigned identity.
+
+After the cluster is created to perform the following steps to obtain the principal id:
```
aramase: After the cluster is created, run these commands to retrieve the principal id

docs/readmes/README.msi.md
```diff
@@ -0,0 +1,50 @@
...
+for VMAs:
```
aramase: nit: VMAs -> VMAS

pkg/cloudprovider/cloudprovider.go
```diff
@@ -59,6 +60,8 @@ func NewCloudProvider(configFile string) (c *Client, e error) {
 		azureConfig.SubscriptionID = os.Getenv("SUBSCRIPTION_ID")
 		azureConfig.ResourceGroupName = os.Getenv("RESOURCE_GROUP")
 		azureConfig.VMType = os.Getenv("VM_TYPE")
+		azureConfig.UseManagedIdentityExtension = strings.EqualFold(os.Getenv("USE_MSI"), "True")
+		azureConfig.UserAssignedIdentityID = os.Getenv("USER_ASSIGNED_MSI_CLIENTID")
```
aramase: I think we should split `CLIENTID` into `CLIENT_ID` to stay in sync with the convention we already have.

pkg/nmi/server/server.go
```diff
@@ -216,6 +219,38 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
...
+		token, err = auth.GetServicePrincipalTokenFromMSIWithUserAssignedID(rqClientID, rqResource)
+	}
+	if err != nil {
+		logger.Errorf("Failed to get service principal token. Error: %+v", err)
+		// TODO: return the right status code based on the error we got from adal.
+		return nil, http.StatusForbidden, err
+	}
+	response, err := json.Marshal(*token)
+	if err != nil {
+		logger.Errorf("Failed to unmarshal service principal token. Error: %+v", err)
```
aramase: Failed to marshal service principal token

pkg/nmi/server/server.go
```diff
@@ -216,6 +219,38 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
...
+		return nil, http.StatusInternalServerError, err
+	}
+	return response, 0, nil
```
aramase: I don't think 0 is a valid http status code. We should return maybe 200 OK, because request was successful.

kkmsft: For success case, the return code is not used. We just write the response. In any case have changed it to http.StatusOK

pkg/nmi/server/server.go
kkmsft: Thanks Anish. Done.

pkg/cloudprovider/cloudprovider.go
kkmsft: done

docs/readmes/README.msi.md
kkmsft: Done.

docs/readmes/README.msi.md
kkmsft: Done

docs/readmes/README.msi.md
kkmsft: Done

PR #266, opened 2019-06-27
Fix the nil cast on error for podlist and refactor pod retry call

kkmsft:
In pod retry the object returned from listing was used to cast even when error was returned. This PR
fixes this issue. The retry call has been refactored to account for other error conditions and simplify
the usage. This PR introduces getPodList call and then a retry call built on top of that.

**Reason for Change**:

Fix the nil cast panic in case of error return from Listpod.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -15,6 +15,7 @@ import (
 
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
 	crd "github.com/Azure/aad-pod-identity/pkg/crd"
+	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/k8s/client.go
```diff
@@ -100,19 +101,53 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
-	if err != nil || len(podList.(*v1.PodList).Items) == 0 {
-		if tries > 1 {
+	if err != nil {
+		return nil, err
+	}
+
+	// Confirm that we are able to cast properly.
+	podList, ok := listObject.(*v1.PodList)
+	if !ok {
+		return nil, fmt.Errorf("list object could not be converted to podllist")
```
aramase: s/podllist/podlist

pkg/k8s/client.go
```diff
@@ -100,19 +101,53 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
+	}
+
+	if podList == nil {
+		return nil, fmt.Errorf("Podlist nil")
```
aramase: nit: errors should start with lower case

pkg/k8s/client.go
```diff
@@ -100,19 +101,53 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
+	}
+
+	if len(podList.Items) == 0 {
+		return nil, fmt.Errorf("Pod List empty")
```
aramase: nit: errors should start with lower case

pkg/k8s/client.go
```diff
@@ -100,19 +101,53 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
+		} else {
+			break
 		}
 	}
-
-	return podList.(*v1.PodList), err
+	if err != nil {
+		return nil, err
+	}
+	if podList == nil {
```
aramase: I don't think this check is necessary as we already check if list is nil in `getPodList`

pkg/k8s/client.go
```diff
@@ -100,19 +101,53 @@ func (c *KubeClient) GetPodName(podip string) (podns, poddname string, err error
...
+	return podList, nil
+}
+
+func (c *KubeClient) getPodListWithTries(podip string, tries int, sleeptime time.Duration) (*v1.PodList, error) {
+	var podList *v1.PodList
+	var err error
+
+	for i := 0; i < tries; i++ {
+		podList, err = c.getPodList(podip)
+		if err != nil {
```
aramase: can we simplify this to be something like this -

```go
	for i := 0; i < tries; i++ {
		podList, err = c.getPodList(podip)
                if err==nil {
                      break
                }

	        log.Warningf("List pod error: %+v. Retrying, attempt number: %d", err, i)
	        time.Sleep(sleeptime * time.Millisecond)
	}
        return podList, err
```

1. check to see if podList is nil is already done in `getPodList`. So if it's nil we'll get an error.

pkg/k8s/client.go
kkmsft: Done

pkg/k8s/client.go
kkmsft: Done

pkg/k8s/client.go
kkmsft: refactored

pkg/k8s/client.go
kkmsft: refactored for simplicity and changed the name to reflect the retry functionality.

kkmsft:
Ran 10 of 10 Specs in 2306.736 seconds

PR #267, opened 2019-06-28
Add custom user agent identifier to the adal calls and update sdk/autorest libraries

kkmsft:
**Reason for Change**:

Adal library calls provides the facility to add customer user agent header to the calls. MIC makes the vm/vmss specific calls and NMI makes calls to get token from IMDS. Custom user agent headers have been added to these calls.

The azure sdk and adal libraries are updated to the latest version available currently.

**Issue Fixed**:

**Notes for Reviewers**:


kkmsft:
Ran 9 of 9 Specs in 1859.396 seconds
SUCCESS! -- 9 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (1859.40s)

version/version.go
```diff
@@ -17,6 +17,13 @@ var MICVersion string
 // NMIVersion is the version of the NMI component
 var NMIVersion string
 
+// GetUserAgent is used to get the user agent string which is then provided to adal
+// to use as the extended user agent header.
+// The format is: aad-pod-identity/<component, either NMI or MIC>/<Version of MIC>/<Version of NMI>/<Git commit>/<Build date>
+func GetUserAgent(component string) string {
+	return fmt.Sprintf("aad-pod-identity/%s/%s/%s/%s/%s", component, MICVersion, NMIVersion, GitCommit, BuildDate)
```
cpuguy83: This ua string seems strange.
Why not pass the version to use as a function argument, or store it in a map as a package var?

Issue #269, opened 2019-06-28
Handle case sensitivity in resourceID format


aramase:
**Describe the bug**
When identity is manually assigned to node using az cli, the resourceID uses `resourceGroups` in the format `/subscriptions/[REDACTED]/resourceGroups/devcluster/providers/Microsoft.ManagedIdentity/userAssignedIdentities/keyvault-identity` . When a new identity is created the resourceID returned however has `resourcegroups` `/subscriptions/[REDACTED]/resourcegroups/devcluster/providers/Microsoft.ManagedIdentity/userAssignedIdentities/keyvault-identity`. We need to normalize the resourceID format before performing a check to see if they are equal.

**Steps To Reproduce**

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**

**Additional context**

kkmsft:
Closing it via #271

PR #271, opened 2019-07-01
Handle case sensitive id check

aramase:
**Reason for Change**:

MIC fails when trying to compare same identity but with different cases for resource groups `resourceGroups` | `resourcegroups` in the resourceID format. This results in error when trying to assign the identity which already exists. This change will ensure we validate the identities case insensitively.

**Issue Fixed**:

https://github.com/Azure/aad-pod-identity/issues/269

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/identity.go
+++ b/pkg/cloudprovider/identity.go
@@ -2,6 +2,7 @@ package cloudprovider
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
 )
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/identity.go
```diff
@@ -65,7 +66,7 @@ func filter(ls *[]string, filter string) {
 	}
 
 	for i, v := range *ls {
-		if v == filter {
+		if strings.EqualFold(v, filter) {
```
aramase: One thing to note - before the identity was never removed if it was manually created and the resourceID format was different in azure identity. With this change it'll be deleted even if the case is different in resourceID.

Issue #257, opened 2019-06-21
Retries in NMI and return 404 when assigned identity not found


aramase:
- Retry in quick succession to get the azure assigned identity
- Retry to get the token in addition to client side retries
- Return 404 instead of 403 when AzureAssginedIdentity is not found after retries

PR #272, opened 2019-07-02
NMI retries and ticker for periodic sync reconcile

aramase:
**Reason for Change**:

- Wrap list pod ids call with retries
- Filter assigned ids based on state in NMI
- Change the status code to `404` instead of `403` when no valid assigned ids found after retries. 
- Add a ticker to periodically run sync cycle to reconcile any crds that are in error state. This is required because if no pod events are generated, then the sync loop will never get triggered to resync.
- Make the sync retry interval configurable

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/257

**Notes for Reviewers**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ e2e:
 
 .PHONY: unit-test
 unit-test:
-	go test $(shell go list ./... | grep -v /test/e2e) -v
+	go test -count=1 $(shell go list ./... | grep -v /test/e2e) -v
 
 .PHONY: validate-version
 validate-version: validate-version-NMI validate-version-MIC validate-version-IDENTITY_VALIDATOR validate-version-DEMO
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"os"
+	"time"
 
 	"github.com/Azure/aad-pod-identity/pkg/mic"
 	"github.com/Azure/aad-pod-identity/version"
@@ -12,10 +13,11 @@ import (
 )
 
 var (
-	kubeconfig      string
-	cloudconfig     string
-	forceNamespaced bool
-	versionInfo     bool
+	kubeconfig        string
+	cloudconfig       string
+	forceNamespaced   bool
+	versionInfo       bool
+	syncRetryInterval string
 )
 
 func main() {
@@ -24,6 +26,8 @@ func main() {
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
+	flag.StringVar(&syncRetryInterval, "syncRetryInterval", "3600", "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
+
 	flag.Parse()
 	if versionInfo {
 		version.PrintVersionAndExit()
@@ -43,7 +47,13 @@ func main() {
 	}
 
 	forceNamespaced = forceNamespaced || "true" == os.Getenv("FORCENAMESPACED")
-	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced)
+
+	syncRetryDuration, err := time.ParseDuration(syncRetryInterval)
+	if err != nil {
+		glog.Fatalf("Could not read syncRetryInterval. Error %+v", err)
+	}
+
+	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration*time.Second)
 	if err != nil {
 		glog.Fatalf("Could not get the MIC client: %+v", err)
 	}
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -53,7 +53,7 @@ func main() {
 		glog.Fatalf("Could not read syncRetryInterval. Error %+v", err)
 	}
 
-	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration*time.Second)
+	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration)
 	if err != nil {
 		glog.Fatalf("Could not get the MIC client: %+v", err)
 	}
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -17,7 +17,7 @@ var (
 	cloudconfig       string
 	forceNamespaced   bool
 	versionInfo       bool
-	syncRetryInterval string
+	syncRetryDuration time.Duration
 )
 
 func main() {
@@ -26,7 +26,7 @@ func main() {
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
-	flag.StringVar(&syncRetryInterval, "syncRetryInterval", "3600s", "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
+	flag.DurationVar(&syncRetryDuration, "syncRetryDuration", 3600*time.Second, "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
 
 	flag.Parse()
 	if versionInfo {
@@ -47,11 +47,6 @@ func main() {
 	}
 
 	forceNamespaced = forceNamespaced || "true" == os.Getenv("FORCENAMESPACED")
-
-	syncRetryDuration, err := time.ParseDuration(syncRetryInterval)
-	if err != nil {
-		glog.Fatalf("Could not read syncRetryInterval. Error %+v", err)
-	}
 
 	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration)
 	if err != nil {
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -26,7 +26,7 @@ func main() {
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
-	flag.StringVar(&syncRetryInterval, "syncRetryInterval", "3600", "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
+	flag.StringVar(&syncRetryInterval, "syncRetryInterval", "3600s", "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
 
 	flag.Parse()
 	if versionInfo {
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -38,7 +38,7 @@ type ClientInt interface {
 	ListBindings() (res *[]aadpodid.AzureIdentityBinding, err error)
 	ListAssignedIDs() (res *[]aadpodid.AzureAssignedIdentity, err error)
 	ListIds() (res *[]aadpodid.AzureIdentity, err error)
-	ListPodIds(podns, podname string, assignedIDStates []string) (*[]aadpodid.AzureIdentity, error)
+	ListPodIds(podns, podname string) (map[string][]aadpodid.AzureIdentity, error)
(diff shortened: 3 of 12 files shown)


pkg/crd/crd.go
```diff
@@ -267,7 +267,7 @@ func (c *Client) ListPodIds(podns, podname string) (*[]aadpodid.AzureIdentity, e
 
 	var matchedIds []aadpodid.AzureIdentity
 	for _, v := range azAssignedIDList.(*aadpodid.AzureAssignedIdentityList).Items {
-		if v.Spec.Pod == podname && v.Spec.PodNamespace == podns {
+		if v.Spec.Pod == podname && v.Spec.PodNamespace == podns && v.Status.Status == "Assigned" {
```
kkmsft: Can we make these states a constant ?

pkg/mic/mic.go
```diff
@@ -152,20 +154,27 @@ func (c *Client) Sync(exit <-chan struct{}) {
...
+			glog.V(6).Infof("Received event: %v", event)
+		case <-ticker.C:
+			glog.V(6).Infof("Running sync retry loop")
 		}
 
 		stats.Init()
 		// This is the only place where the AzureAssignedIdentity creation is initiated.
 		begin := time.Now()
 		workDone := false
 		glog.V(6).Infof("Received event: %v", event)
```
kkmsft: What will event show here ?

pkg/nmi/server/server.go
```diff
@@ -232,14 +237,15 @@ func (s *Server) msiHandler(logger *log.Entry, w http.ResponseWriter, r *http.Re
...
 	}
-	podIDs, err := s.KubeClient.ListPodIds(podns, podname)
-	if err != nil || len(*podIDs) == 0 {
+
+	podIDs, err := listPodIDsWithRetry(s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttempts)
+	if err != nil {
 		msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s", podns, podname)
 		logger.Errorf("%s, %+v", msg, err)
-		http.Error(w, msg, http.StatusForbidden)
+		http.Error(w, msg, http.StatusNotFound)
```
kkmsft: hosthandler and mishandler looks identical through most parts.

pkg/nmi/server/server.go
```diff
@@ -380,3 +386,29 @@ func handleTermination() {
...
+				if strings.EqualFold(rqClientID, podID.Spec.ClientID) {
+					return podIDs, nil
+				}
+			}
+		}
+		attempt++
+		logger.Warningf("failed to get assigned ids for pod:%s/%s, retrying attempt: %d", podns, podname, attempt)
+		time.Sleep(listPodIDsRetryIntervalInSeconds * time.Second)
+	}
+	return nil, fmt.Errorf("getting assigned identities for pod %s/%s failed after %d attempts. Error: %v", podns, podname, maxAttempts, err)
```
kkmsft: will this error propagate to the caller in the application ?

pkg/nmi/server/server.go
```diff
@@ -380,3 +386,29 @@ func handleTermination() {
...
+			// exists and is in Assigned state
+			for _, podID := range *podIDs {
+				if strings.EqualFold(rqClientID, podID.Spec.ClientID) {
+					return podIDs, nil
+				}
+			}
+		}
+		attempt++
+		logger.Warningf("failed to get assigned ids for pod:%s/%s, retrying attempt: %d", podns, podname, attempt)
+		time.Sleep(listPodIDsRetryIntervalInSeconds * time.Second)
```
kkmsft: Where do we break if we go >= attempt

pkg/nmi/server/server.go
```diff
@@ -380,3 +386,29 @@ func handleTermination() {
...
+
+func listPodIDsWithRetry(kubeClient k8s.Client, logger *log.Entry, podns, podname, rqClientID string, maxAttempts int) (*[]aadpodid.AzureIdentity, error) {
+	attempt := 0
+	var err error
+	var podIDs *[]aadpodid.AzureIdentity
+
+	for {
+		podIDs, err = kubeClient.ListPodIds(podns, podname)
+		if err == nil && len(*podIDs) != 0 {
+			if len(rqClientID) == 0 {
```
kkmsft: there are two kinds of error, one is that after a few seconds, we still did not find the AzureAssignedIdentity and another is that it never moved to Assigned. Not able to gather that different from retry attempts and varying time durations for them..?

pkg/nmi/server/server.go
aramase: Yeah, this will be propagated if it fails after retries.

pkg/crd/crd.go
aramase: That was my initial plan, but the only confusion is IdentityCreated is an int (EventType) in the v1 pkg.

pkg/nmi/server/server.go
aramase: The retry attempts are not different based on the states. The final desired state is `Assigned`. So I think this encompasses all possible scenarios.

aramase:
Ran 10 of 10 Specs in 4998.515 seconds
--- PASS: TestAADPodIdentity (4998.52s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	4998.532s

test/e2e/aadpodidentity_test.go
```diff
@@ -111,7 +111,7 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 		fmt.Println("\nTearing down the test environment...")
 
 		// Ensure a clean cluster after the end of each test
-		cmd := exec.Command("kubectl", "delete", "AzureIdentity,AzureIdentityBinding,AzureAssignedIdentity", "--all")
+		cmd := exec.Command("kubectl", "delete", "AzureIdentity,AzureIdentityBinding", "--all")
```
aramase: Removing azureassignedidentity from the delete command. It should be deleted by MIC once the identity and binding are removed and it's a good test to ensure we reliably delete the assigned identities.

Makefile
```diff
@@ -141,7 +141,7 @@ e2e:
 
 .PHONY: unit-test
 unit-test:
-	go test $(shell go list ./... | grep -v /test/e2e) -v
+	go test -count=1 $(shell go list ./... | grep -v /test/e2e) -v
```
kkmsft: What is the usage of count here ?

pkg/crd/crd.go
```diff
@@ -268,7 +268,11 @@ func (c *Client) ListPodIds(podns, podname string) (*[]aadpodid.AzureIdentity, e
 	var matchedIds []aadpodid.AzureIdentity
 	for _, v := range azAssignedIDList.(*aadpodid.AzureAssignedIdentityList).Items {
 		if v.Spec.Pod == podname && v.Spec.PodNamespace == podns {
-			matchedIds = append(matchedIds, *v.Spec.AzureIdentityRef)
+			for _, desiredState := range assignedIDStates {
+				if v.Status.Status == desiredState {
```
kkmsft: strings.EqualFold ? or compareState function which we can change as per version ?

pkg/nmi/server/server.go
kkmsft: Can you please unify the new code which you are adding into one single function.

pkg/nmi/server/server.go
```diff
@@ -380,3 +409,46 @@ func handleTermination() {
 	log.Infof("Exiting with %v", exitCode)
 	os.Exit(exitCode)
 }
+
+func listPodIDsInCreatedStateWithRetries(ctx context.Context, kubeClient k8s.Client, logger *log.Entry, podns, podname, rqClientID string, maxAttempts int) (*[]aadpodid.AzureIdentity, error) {
+	// created and assigned states are added to statesToCheck. This is to ensure if the call is made after the assigned identity is in assigned state
+	// for the pod, we shouldn't fail on the hard check for created state. Both states are accepted values for passing first check.
+	return listPodIDsWithRetry(ctx, kubeClient, logger, podns, podname, rqClientID, []string{aadpodid.AssignedIDCreated, aadpodid.AssignedIDAssigned}, maxAttempts)
```
kkmsft: Why are both state assigned ids added when the function name indicates we are going to gather Created ones.

pkg/nmi/server/server.go
```diff
@@ -167,15 +173,25 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
...
-	if err != nil || len(*podIDs) == 0 {
-		msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s", podns, podname)
+	// this check will ensure assigned id exists with created/assigned state
+	// no error on this check means we can proceed with the second check to ensure assigned identity with assigned state exists
+	_, err := listPodIDsInCreatedStateWithRetries(r.Context(), s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttempts)
+	if err != nil {
+		msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s in created state", podns, podname)
 		logger.Errorf("%s, %+v", msg, err)
-		http.Error(w, msg, http.StatusForbidden)
+		http.Error(w, msg, http.StatusNotFound)
```
kkmsft: Why do we have to do it two times ? Shouldn't we check and do accordingly ?

test/e2e/aadpodidentity_test.go
kkmsft: if that is the exception we should add the check at the end to see if no AzureAssignedIdentities exists and fail the test if it does.

aramase: We already ensure the length of azureassignedidentities is equal to 0 at the end of `AfterEach`.

Makefile
aramase: It's to ensure the test runs at least once instead of showing the cached result.

pkg/nmi/server/server.go
aramase: I'll add this as a TODO and refactor in the next PR. Have a few questions on why we check filter identities using force namespaced only in 1 handler and not another. We can go over that during the review.

cmd/mic/main.go
```diff
@@ -24,6 +26,8 @@ func main() {
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
+	flag.StringVar(&syncRetryInterval, "syncRetryInterval", "3600s", "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
```
aramase: Change to duration var

aramase:
```bash
Ran 11 of 11 Specs in 3839.814 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (3839.81s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	3839.837s
```

pkg/crd/crd.go
```diff
@@ -257,22 +257,22 @@ func (c *Client) ListIds() (res *[]aadpodid.AzureIdentity, err error) {
...
 	if err != nil {
 		glog.Error(err)
 		return nil, err
 	}
 
-	var matchedIds []aadpodid.AzureIdentity
+	idStateMap := make(map[string][]aadpodid.AzureIdentity)
 	for _, v := range azAssignedIDList.(*aadpodid.AzureAssignedIdentityList).Items {
 		if v.Spec.Pod == podname && v.Spec.PodNamespace == podns {
-			matchedIds = append(matchedIds, *v.Spec.AzureIdentityRef)
```
kkmsft: Can we please add a comment in the next PR indicating the backward compatibility aspects.

test/e2e/template/deployment-rbac.yaml
```diff
@@ -85,6 +85,8 @@ metadata:
   name: nmi
   namespace: {{.Namespace}}
 spec:
+  updateStrategy:
```
kkmsft: What is the significance of this ?

pkg/nmi/server/server.go
```diff
@@ -380,3 +384,57 @@ func handleTermination() {
...
+
+		select {
+		case <-time.After(listPodIDsRetryIntervalInSeconds * time.Second):
+		case <-ctx.Done():
+			err = ctx.Err()
+			return nil, err
+		}
+		logger.Warningf("failed to get assigned ids for pod:%s/%s in ASSIGNED state, retrying attempt: %d", podns, podname, attempt)
+	}
+	return nil, fmt.Errorf("getting assigned identities for pod %s/%s in ASSIGNED state failed after %d attempts. Error: %v", podns, podname, 2*maxAttemptsPerCheck, err)
```
kkmsft: It would be good to have the duration in the error logs on how much it tried to do retry.

PR #277, opened 2019-07-05
Add leader election code to MIC

kkmsft:
**Reason for Change**:

Add leader election code to MIC to run it as ReplicaSet with multiple replicas which can come up and continue in case of a failure of the active MIC. Also added are deployment yamls which configures the ReplicaSet appropriately under deploy/master/replicaset.

**Issue Fixed**:

**Notes for Reviewers**:

TODO: 
1. Add tests in e2e

### DIFF
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -39,7 +39,7 @@ func main() {
 	flag.StringVar(&leaderElectionCfg.ThisLeaderName, "leader-election-name", hostName, "leader name. default is 'hostname'")
 	flag.StringVar(&leaderElectionCfg.LeaderElectionNamespace, "leader-election-namespace", "default", "name space to create leader election objects. default is 'default' namesapce")
 	flag.StringVar(&leaderElectionCfg.LeaderElectionId, "leader-election-id", "aad-pod-identity-mic", "leader election id")
-	flag.DurationVar(&leaderElectionCfg.LeaderElectionTtl, "leader-election-ttl", time.Second*30, "leader election ttl")
+	flag.DurationVar(&leaderElectionCfg.LeaderElectionTtl, "leader-election-ttl", time.Second*15, "leader election ttl")
 
 	flag.Parse()
 	if versionInfo {
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -18,15 +18,28 @@ var (
 	forceNamespaced   bool
 	versionInfo       bool
 	syncRetryDuration time.Duration
+	leaderElectionCfg mic.LeaderElectionConfig
 )
 
 func main() {
 	defer glog.Flush()
+	hostName, err := os.Hostname()
+	if err != nil {
+		glog.Errorf("Get hostname failure.")
+		glog.Flush()
+		os.Exit(1)
+	}
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
 	flag.DurationVar(&syncRetryDuration, "syncRetryDuration", 3600*time.Second, "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
+
+	// Leader election parameters
+	flag.StringVar(&leaderElectionCfg.ThisLeaderName, "leader-election-name", hostName, "leader name. default is 'hostname'")
+	flag.StringVar(&leaderElectionCfg.LeaderElectionNamespace, "leader-election-namespace", "default", "name space to create leader election objects. default is 'default' namesapce")
+	flag.StringVar(&leaderElectionCfg.LeaderElectionId, "leader-election-id", "aad-pod-identity-mic", "leader election id")
+	flag.DurationVar(&leaderElectionCfg.LeaderElectionTtl, "leader-election-ttl", time.Second*30, "leader election ttl")
 
 	flag.Parse()
 	if versionInfo {
@@ -49,6 +62,7 @@ func main() {
 	forceNamespaced = forceNamespaced || "true" == os.Getenv("FORCENAMESPACED")
 
 	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration)
+	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, &leaderElectionCfg)
 	if err != nil {
 		glog.Fatalf("Could not get the MIC client: %+v", err)
 	}
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -25,7 +25,7 @@ func main() {
 	defer glog.Flush()
 	hostName, err := os.Hostname()
 	if err != nil {
-		glog.Fatalf("Get hostname failure.")
+		glog.Fatalf("Get hostname failure. Error: %+v", err)
 	}
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
@@ -34,9 +34,9 @@ func main() {
 	flag.DurationVar(&syncRetryDuration, "syncRetryDuration", 3600*time.Second, "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
 
 	// Leader election parameters
-	flag.StringVar(&leaderElectionCfg.ResourceName, "leader-election-name", hostName, "leader name. default is 'hostname'")
-	flag.StringVar(&leaderElectionCfg.Namespace, "leader-election-namespace", "default", "name space to create leader election objects. default is 'default' namesapce")
-	flag.StringVar(&leaderElectionCfg.ID, "leader-election-id", "aad-pod-identity-mic", "leader election id")
+	flag.StringVar(&leaderElectionCfg.Instance, "leader-election-instance", hostName, "leader election instance name. default is 'hostname'")
+	flag.StringVar(&leaderElectionCfg.Namespace, "leader-election-namespace", "default", "namespace to create leader election objects")
+	flag.StringVar(&leaderElectionCfg.Name, "leader-election-name", "aad-pod-identity-mic", "leader election name")
 	flag.DurationVar(&leaderElectionCfg.Duration, "leader-election-duration", time.Second*15, "leader election duration")
 
 	flag.Parse()
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -25,9 +25,7 @@ func main() {
 	defer glog.Flush()
 	hostName, err := os.Hostname()
 	if err != nil {
-		glog.Errorf("Get hostname failure.")
-		glog.Flush()
-		os.Exit(1)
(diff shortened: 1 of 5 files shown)


kkmsft:
Ran 10 of 10 Specs in 2939.239 seconds
--- PASS: TestAADPodIdentity (2939.24s)

cmd/mic/main.go
```diff
@@ -18,16 +18,27 @@ var (
...
 	versionInfo       bool
 	syncRetryDuration time.Duration
+	leaderElectionCfg mic.LeaderElectionConfig
 )
 
 func main() {
 	defer glog.Flush()
+	hostName, err := os.Hostname()
+	if err != nil {
+		glog.Fatalf("Get hostname failure.")
```
aramase: Need to include the error in the err msg

cmd/mic/main.go
```diff
@@ -18,16 +18,27 @@ var (
...
+	}
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
 	flag.DurationVar(&syncRetryDuration, "syncRetryDuration", 3600*time.Second, "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
 
+	// Leader election parameters
+	flag.StringVar(&leaderElectionCfg.ResourceName, "leader-election-name", hostName, "leader name. default is 'hostname'")
+	flag.StringVar(&leaderElectionCfg.Namespace, "leader-election-namespace", "default", "name space to create leader election objects. default is 'default' namesapce")
```
aramase: We don't need to explicitly say the default value in the usage. Since the default value is defined, pflag will show that in the output.

For instance - 
```
  -syncRetryDuration duration
    	The interval in seconds at which sync loop should periodically check for errors and reconcile. (default 1h0m0s)
```

deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -0,0 +1,208 @@
...
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        component: mic
+    spec:
+      serviceAccountName: aad-pod-id-mic-service-account
+      containers:
+      - name: mic
+        image: "mcr.microsoft.com/k8s/aad-pod-identity/mic:1.3"
```
aramase: Image should be `1.5-rc1` image here instead of `1.3`

deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -0,0 +1,208 @@
...
+      serviceAccountName: aad-pod-id-nmi-service-account
+      hostNetwork: true
+      volumes:
+      - hostPath:
+          path: /run/xtables.lock
+          type: FileOrCreate
+        name: iptableslock
+      containers:
+      - name: nmi
+        image: "mcr.microsoft.com/k8s/aad-pod-identity/nmi:1.4"
```
aramase: Image should be `1.5-rc1` instead of `1.4`

deploy/infra/master/replicaset/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+    spec:
+      hostNetwork: true
+      volumes:
+      - hostPath:
+          path: /run/xtables.lock
+          type: FileOrCreate
+        name: iptableslock
+      containers:
+      - name: nmi
+        image: "mcr.microsoft.com/k8s/aad-pod-identity/nmi:master"
```
aramase: same as above

deploy/infra/master/replicaset/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+spec:
+  replicas: 2
+  template:
+    metadata:
+      labels:
+        component: mic
+    spec:
+      containers:
+      - name: mic
+        image: "mcr.microsoft.com/k8s/aad-pod-identity/mic:master"
```
aramase: same as above

pkg/mic/mic.go
```diff
@@ -99,7 +112,58 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
...
+	c.LeaderElectionConfig = leaderElectionConfig
+	resourceLock, err := resourcelock.New(resourcelock.EndpointsResourceLock,
+		c.Namespace,
+		c.ID,
+		clientSet.CoreV1(),
+		resourcelock.ResourceLockConfig{
+			Identity:      c.ResourceName,
+			EventRecorder: recorder})
+	if err != nil {
+		glog.Errorf("Resource lock creation for leadeer election failed with error : %v", err)
```
aramase: s/leadeer/leader

pkg/mic/mic.go
```diff
@@ -99,7 +112,58 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
...
+				glog.Errorf("Lost Leader Lease")
+				glog.Flush()
+				os.Exit(1)
+			},
+		},
+		Lock: resourceLock,
+	}
+
+	leaderElector, err = leaderelection.NewLeaderElector(config)
+	if err != nil {
```
aramase: We can skip the err != nil check here and return from here

test/e2e/template/deployment-rbac.yaml
```diff
@@ -140,6 +140,9 @@ rules:
 - apiGroups: [""]
   resources: ["events"]
   verbs: ["create", "patch"]
+- apiGroups: [""]
+  resources: ["endpoints"]
+  verbs: ["create", "get", "list", "watch", "update", "patch"]
```
aramase: Why do we need `create`, `update` and `patch`?

cmd/mic/main.go
```diff
@@ -18,16 +18,27 @@ var (
...
+		glog.Fatalf("Get hostname failure.")
+	}
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to the kube config")
 	flag.StringVar(&cloudconfig, "cloudconfig", "", "Path to cloud config e.g. Azure.json file")
 	flag.BoolVar(&forceNamespaced, "forceNamespaced", false, "Forces namespaced identities, binding, and assignment")
 	flag.BoolVar(&versionInfo, "version", false, "Prints the version information")
 	flag.DurationVar(&syncRetryDuration, "syncRetryDuration", 3600*time.Second, "The interval in seconds at which sync loop should periodically check for errors and reconcile.")
 
+	// Leader election parameters
+	flag.StringVar(&leaderElectionCfg.ResourceName, "leader-election-name", hostName, "leader name. default is 'hostname'")
```
khenidak: default shouldn't be host name, default should be a well known string. Mic pods will get different host names.

cmd/mic/main.go
khenidak: nit `namespace` not `name space`

pkg/mic/mic.go
```diff
@@ -99,7 +112,58 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
 		NodeClient:        &NodeClient{informer.Core().V1().Nodes()},
 		IsNamespaced:      isNamespaced,
 		syncRetryInterval: syncRetryInterval,
-	}, nil
+	}
+	leaderElector, err := c.NewLeaderElector(clientSet, recorder, leaderElectionConfig)
+	if err != nil {
+		return nil, err
```
khenidak: please log here.

test/e2e/template/deployment-rbac.yaml
khenidak: hmm we need `create` and `get` for sure, the rest i am not so sure about.

pkg/mic/mic.go
```diff
@@ -99,7 +112,58 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
...
+		RenewDeadline: c.Duration / 2,
+		RetryPeriod:   c.Duration / 4,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(exit <-chan struct{}) {
+				c.Start(exit)
+			},
+			OnStoppedLeading: func() {
+				glog.Errorf("Lost Leader Lease")
+				glog.Flush()
+				os.Exit(1)
```
khenidak: nit the case of log line consistent. don't use title case, and generally avoid caps all together.

deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -0,0 +1,208 @@
+apiVersion: v1
```
khenidak: Why do we have these new files, are they new?

cmd/mic/main.go
kkmsft: isn't that for ID ? Shouldn't the resource name be unique ?

deploy/infra/master/replicaset/deployment-rbac.yaml
kkmsft: right now the deployment file (https://github.com/Azure/aad-pod-identity/blob/master/deploy/infra/deployment-rbac.yaml) under the infra has the extra parameter with the binary name, because of a bug in the Docker packaging which we fixed later. That file needs to be remain as is so that anyone using the stable images can use that deployment file. These file contains the right yaml for the images from the master. When we do next release - 1.5 this file will move to (https://github.com/Azure/aad-pod-identity/blob/master/deploy/infra/deployment-rbac.yaml) so that stable images can be obtained by using the same deployment files and also have replicaset configuration.

pkg/mic/mic.go
kkmsft: done

pkg/mic/mic.go
kkmsft: Changed. Only first one is caps now.

cmd/mic/main.go
khenidak: `leaderElectionCfg.ResourceName` throw me off. i thought it is for endpoint name, not for the identity of the leader. Let us try to avoid the same for future users. 
- rename the field ResourceName to leader identity or something of that sort. 
- rename field `ID` to `Identity`

cmd/mic/main.go
kkmsft: Done

kkmsft: done.

pkg/mic/mic.go
kkmsft: done

cmd/mic/main.go
kkmsft: Done.

cmd/mic/main.go
kkmsft: ResourceName ==> Instance
ID ==> Name

kkmsft:
Ran 12 of 12 Specs in 2723.302 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (2723.30s)
PASS
ok      github.com/Azure/aad-pod-identity/test/e2e      2723.340s

deploy/infra/master/replicaset/deployment.yaml
aramase: @kkmsft looks like the image version is still master here?

deploy/infra/master/replicaset/deployment.yaml
aramase: @kkmsft looks like the image version is still master here instead of rc?

PR #292, opened 2019-07-15
Update error status code based on state

aramase:
**Reason for Change**:

Retries have been added in NMI as part of my previous PR - https://github.com/Azure/aad-pod-identity/pull/272. Instead of returning a 404 after internal retries in all scenarios we need to determine the state and then respond with different status codes.
- If no assigned identity found in CREATED state for the request, then we will return `403` which is non-retriable status code in go-autorest adal lib. This means the pod is not part of the current sync cycle and will be processed as part of next sync cycle. By returning a 403 we ensure the retry isn't aggressive.
- If assigned identity found in CREATED state, but not found in ASSIGNED state, then we return a 404 which is a retriable status code. The CREATED state for the identity means, the pod is part of this sync cycle and MIC is in the process of assigning the identity to the node. By returning a 404 the client can retry immediately to request the token. 

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/nmi/server/server.go
+++ b/pkg/nmi/server/server.go
@@ -174,11 +174,23 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
 		http.Error(w, "missing 'podname' and 'podns' from request header", http.StatusBadRequest)
 		return
 	}
-	podIDs, err := listPodIDsWithRetry(r.Context(), s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttempts)
+	podIDs, identityInCreatedStateFound, err := listPodIDsWithRetry(r.Context(), s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttempts)
 	if err != nil {
 		msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s in assigned state", podns, podname)
 		logger.Errorf("%s, %+v", msg, err)
-		http.Error(w, msg, http.StatusNotFound)
+		var statusCode int
+		// if at least an identity was found in created state then we return 404 which is a retriable error code
+		// in the go-autorest library. If the identity is in CREATED state then the identity is being processed in
+		// this sync cycle and should move to ASSIGNED state soon.
+		if identityInCreatedStateFound {
+			statusCode = http.StatusNotFound
+		} else {
+			// if no identity in at least CREATED state was found, then it means the identity creation is not part of the
+			// current ongoing sync cycle. So we return 403 which a non-retriable error code so we give mic enough time to
+			// finish current sync cycle and process identity in the next sync cycle.
+			statusCode = http.StatusForbidden
+		}
+		http.Error(w, msg, statusCode)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


aramase:
```bash
Ran 11 of 11 Specs in 3839.814 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (2313.76s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	2313.784s
```

aramase:
```bash
Ran 11 of 11 Specs in 2337.809 seconds
SUCCESS! -- 11 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (2337.81s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	2337.833s
```

pkg/nmi/server/server.go
```diff
@@ -174,11 +175,23 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
 		http.Error(w, "missing 'podname' and 'podns' from request header", http.StatusBadRequest)
 		return
 	}
-	podIDs, err := listPodIDsWithRetry(r.Context(), s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttempts)
+	podIDs, identityInCreatedStateFound, err := listPodIDsWithRetry(r.Context(), s.KubeClient, logger, podns, podname, rqClientID, listPodIDsRetryAttemptsForCreated, listPodIDsRetryAttemptsForAssigned)
 	if err != nil {
 		msg := fmt.Sprintf("no AzureAssignedIdentity found for pod:%s/%s in assigned state", podns, podname)
 		logger.Errorf("%s, %+v", msg, err)
-		http.Error(w, msg, http.StatusNotFound)
```
kkmsft: Can we please move this to a function, which we can call from both paths

pkg/nmi/server/server.go
```diff
@@ -174,11 +175,23 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
...
+		var statusCode int
+		// if at least an identity was found in created state then we return 404 which is a retriable error code
+		// in the go-autorest library. If the identity is in CREATED state then the identity is being processed in
+		// this sync cycle and should move to ASSIGNED state soon.
+		if identityInCreatedStateFound {
+			statusCode = http.StatusNotFound
+		} else {
+			// if no identity in at least CREATED state was found, then it means the identity creation is not part of the
+			// current ongoing sync cycle. So we return 403 which a non-retriable error code so we give mic enough time to
+			// finish current sync cycle and process identity in the next sync cycle.
```
kkmsft: how is that we give enough time for mic to finish current cycle by passing back a non-retriable error ?

pkg/nmi/server/server.go
```diff
@@ -463,17 +491,18 @@ func listPodIDsWithRetry(ctx context.Context, kubeClient k8s.Client, logger *log
...
 				for _, podID := range idStateMap[aadpodid.AssignedIDAssigned] {
 					if strings.EqualFold(rqClientID, podID.Spec.ClientID) {
-						return idStateMap[aadpodid.AssignedIDAssigned], nil
+						return idStateMap[aadpodid.AssignedIDAssigned], true, nil
 					}
 				}
 			}
-			if len(idStateMap[aadpodid.AssignedIDCreated]) == 0 && attempt >= maxAttemptsPerCheck {
-				return nil, fmt.Errorf("getting assigned identities for pod %s/%s in CREATED state failed after %d attempts. Error: %v", podns, podname, maxAttemptsPerCheck, err)
+			if len(idStateMap[aadpodid.AssignedIDCreated]) == 0 && attempt >= maxAttemptsForCreated {
```
kkmsft: What happens if we move to assigned state before we reach this check ?

pkg/nmi/server/server.go
```diff
@@ -463,17 +491,18 @@ func listPodIDsWithRetry(ctx context.Context, kubeClient k8s.Client, logger *log
...
 					if strings.EqualFold(rqClientID, podID.Spec.ClientID) {
 						logger.Warningf("found assignedIDs with no state for pod:%s/%s. AssignedIDs created with old version of mic.", podns, podname)
-						return idStateMap[""], nil
+						return idStateMap[""], true, nil
 					}
 				}
 				for _, podID := range idStateMap[aadpodid.AssignedIDAssigned] {
 					if strings.EqualFold(rqClientID, podID.Spec.ClientID) {
-						return idStateMap[aadpodid.AssignedIDAssigned], nil
+						return idStateMap[aadpodid.AssignedIDAssigned], true, nil
```
kkmsft: When do we return identities which are in created state ?

pkg/nmi/server/server.go
aramase: The identities are either in assigned state or created state. So during that iteration if its in assigned state, then the check for assigned state first will take care of it. There is no intermittent state possible which will result in assigned and created being empty at same time.

pkg/nmi/server/server.go
aramase: We don't need to return identities which are in created state. We only need identities in Assigned state.

pkg/nmi/server/server.go
aramase: It just means there is no aggressive retry from the go-autorest sdk. Instead of retrying immediately in ~2s which is configured in sdk, now the retry is solely based on the client implementation.

PR #293, opened 2019-07-16
Add e2e test for vmss deletion

cpuguy83:
**Reason for Change**:

Adds an e2e test testing deletion of vmss identities.

**Issue Fixed**:

**Notes for Reviewers**:

I have not run this test yet and am not particularly familiar with Ginkgo. Be kind 👼 🙇 
Posting this for early input.

### DIFF
--- a/test/common/azure/azure.go
+++ b/test/common/azure/azure.go
@@ -9,8 +9,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
 )
+
+type Resource = azure.Resource
 
 // UserAssignedIdentity is used to parse user assigned identity data from 'az vm identity show'
 type UserAssignedIdentity struct {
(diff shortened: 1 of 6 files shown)


cpuguy83:
Ok I've updated this, test is passing... however I wonder how valid this test is if we cannot introspect into mic to figure out if it has done a reconciliation after pod deletion.

cpuguy83:
To run the test, you need a VMSS cluster, and set `AAD_POD_ID_VMSS_NAME` to the name of the vmss, e.g. `export AAD_POD_ID_VMSS_NAME=k8s-linuxpool1-18385292-vmss`... along with all the normal e2e variables.

cpuguy83:
Running into what seems like some flakiness, well keep doing some runs. But I think this is good to go. At the very least e2e at least understands vmss now.

Also added auto-detection for vmss by probing all nodes in the cluster and checking for a vmss with > 1 node.
For the vmss specific test if one is not found the test is ignored.

PR #306, opened 2019-07-24
Marshalling token response in original format

### LINKED ISSUE #275: Backward compatibility broken on update to latest go-autorest sdk
**Describe the bug**
Latest go-autorest sdk version is not backward compatible with the older version. Breaking changes was introduced with this PR - https://github.com/Azure/go-autorest/pull/326. `ExpiresIn`, `ExpiresOn` and `NotAfter` field types have been changed to `json.Number` from `string`. This means if the client is using an older version of go-autorest sdk, then the token returned from the nmi will fail to unmarshal. 

**Steps To Reproduce**
Running e2e tests on master with old version of identity-validator

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**

**Additional context**

### LINKED ISSUE #303: Cluster with MSI - adal: Failed to unmarshal the service principal token during refresh.
**Describe the bug**
Configured AzureIdentityBinding for pod `external-dns` with the new support for user-assigned-identity in the cluster's azure.json file.

99% working - seems to be one last json unmarshalling problem getting the token back to the caller?

**Steps To Reproduce**
Configure external-dns to update cluster DnsZone records using the cluster's managed identity.
Values used for helm deployment:
```
  # https://github.com/helm/charts/tree/master/stable/external-dns
  # https://github.com/helm/charts/tree/master/stable/external-dns/values.yaml
  # https://github.com/kubernetes-incubator/external-dns

  provider: azure

  rbac:
    create: true

  nodeSelector:
    beta.kubernetes.io/os: linux

  securityContext: 
    runAsUser: 0   # container runs as root in order to use cluster's MSI identity
```
After installing aad-pod-identity to cluster - create AzureIdentity and AzureIdentityBinding to associate a user-assigned identity with the pod.

The MIC correctly adds the managed identity to linux agent vm, and the NMI appears to understand the association between the calling pod and the allowed identity, but there is an error being raised at some point during the return path.

An error states apparently valid token response can't be unmarshaled because the expires_in field data time is string... Example error from NMI and the ext-dns logs is below.

**Expected behavior**
The external-dns pod will begin using the MSI token provided by NMI.

**AAD Pod Identity version**
Current commit, `68cc55c7bfee1d5ddc40232c9691900eb9abd5ca`

**Kubernetes version**
1.15.0

**Additional context**
Log lines what happen each time the external-dns pod tried to get an access token. It's unclear which parts of the message are from error in the nmi or the ext-dns process.

```
[nmi-jdpwh] time="2019-07-24T01:38:15Z" level=info msg="matched identityType:0 clientid:7efdc--redacted--cb70 resource:https://management.core.windows.net/" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=10.240.0.6
[external-dns-788c474cc9-zkd5g] time="2019-07-24T01:38:15Z" level=error msg="azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/cd0f--redacted--1353/resourceGroups/redacted-rg/providers/Microsoft.Network/dnsZones?api-version=2016-04-01: StatusCode=0 -- Original Error: adal: Failed to unmarshal the service principal token during refresh. Error = 'json: cannot unmarshal number into Go struct field Token.expires_in of type string' JSON = '{\"access_token\":\"eyJ0eXAi--redacted--gb4KlFL6w\",\"refresh_token\":\"\",\"expires_in\":28800,\"expires_on\":1563958345,\"not_before\":1563929245,\"resource\":\"https://management.core.windows.net/\",\"token_type\":\"Bearer\"}'"
```

lodejard:
Introduces msiResponse struct to marshal token in
a format that matches the underlying metadata
endpoint more closely. The `expires_in`, `expires_on`,
and `not_before` fields are originally provided as
quoted strings.

**Reason for Change**:
Increases compatibility with existing service images
built on older versions of adal client libraries.

**Issue Fixed**:
Fixes #303
Fixes #275

### DIFF
--- a/pkg/nmi/server/server.go
+++ b/pkg/nmi/server/server.go
@@ -220,6 +220,33 @@ func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.R
 	w.Write(response)
 }
 
+// msiResponse marshals in a format that matches the underlying
+// metadata endpoint more closely. This increases compatibility
+// with callers built on older versions of adal client libraries.
+type msiResponse struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+
+	ExpiresIn string `json:"expires_in"`
+	ExpiresOn string `json:"expires_on"`
+	NotBefore string `json:"not_before"`
+
+	Resource string `json:"resource"`
+	Type     string `json:"token_type"`
+}
+
+func newMSIResponse(token adal.Token) msiResponse {
+	return msiResponse{
+		AccessToken:  token.AccessToken,
+		RefreshToken: token.RefreshToken,
+		ExpiresIn:    token.ExpiresIn.String(),
+		ExpiresOn:    token.ExpiresOn.String(),
+		NotBefore:    token.NotBefore.String(),
+		Resource:     token.Resource,
+		Type:         token.Type,
+	}
+}
+
 func (s *Server) isMIC(podNS, rsName string) bool {
 	micRegEx := regexp.MustCompile(`^mic-*`)
 	if strings.EqualFold(podNS, s.MICNamespace) && micRegEx.MatchString(rsName) {
@@ -244,7 +271,7 @@ func (s *Server) getMICToken(logger *log.Entry, rqClientID, rqResource string) (
 		// TODO: return the right status code based on the error we got from adal.
 		return nil, http.StatusForbidden, err
 	}
-	response, err := json.Marshal(*token)
+	response, err := json.Marshal(newMSIResponse(*token))
 	if err != nil {
 		logger.Errorf("Failed to marshal service principal token. Error: %+v", err)
 		return nil, http.StatusInternalServerError, err
@@ -301,7 +328,7 @@ func (s *Server) msiHandler(logger *log.Entry, w http.ResponseWriter, r *http.Re
 		http.Error(w, err.Error(), http.StatusForbidden)
 		return
 	}
-	response, err := json.Marshal(*token)
+	response, err := json.Marshal(newMSIResponse(*token))
 	if err != nil {
 		logger.Errorf("failed to marshal service principal token for pod:%s/%s, %+v", podns, podname, err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)


pkg/nmi/server/server.go
```diff
@@ -244,7 +271,7 @@ func (s *Server) getMICToken(logger *log.Entry, rqClientID, rqResource string) (
 		// TODO: return the right status code based on the error we got from adal.
 		return nil, http.StatusForbidden, err
 	}
-	response, err := json.Marshal(*token)
+	response, err := json.Marshal(newMSIResponse(*token))
```
kkmsft: Is there a way to check the sdk version from which the call was originating and then return back either the new format or the new format. That would make it work for both versions. What do you say ?

aramase: @lodejard Thank you for this PR. We can get the sdk version from the request url. The key name is `api-version`. I think we should make a decision based on the api-version and return the appropriate token response.

aramase: You can use this - https://github.com/Azure/aad-pod-identity/blob/master/pkg/nmi/server/server.go#L369-L376 to also get the `api-version` and then make a decision if we need to send the token as is or create a new MSI response.

lodejard: That's an interesting thought. @aramase had a similar question just below. I'll answer on this thread.

It seems like the most important thing is for the NMI response body to be as nearly identical to the underlying IMDS endpoint as possible. Do we know if the original response body returns unquoted numbers based on the `api-version` on the request? Let's see...

lodejard: It does look like the api specification for even the latest version declares those fields as `string` in the [IdentityTokenResponse](https://github.com/Azure/azure-rest-api-specs/blob/8f2e7ec/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2019-06-01/imds.json#L868).

I suspect the change to `type json.Number string` in the `adal.Token` was more for convenience in end-user code, and they didn't anticipate that type would be used to round-trip Marshal the data back into a response wire-form.

lodejard: (merging thread with above)

lodejard: Confirmed on a VM also - looks like those are actually string fields even on the newest api-version.

```
curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-10-01&resource=https://management.core.windows.net/' --header 'Metadata: true' -D -

HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Server: Microsoft-IIS/10.0
Date: Wed, 24 Jul 2019 22:07:19 GMT
Content-Length: 1683

{"access_token":"eyJ0--(redacted)--b71c","expires_in":"28800","expires_on":"1564028884","ext_expires_in":"28800","not_before":"1563999784","resource":"https://management.core.windows.net/","token_type":"Bearer"}
```

kkmsft: @lodejard - I don't think there was a change in IMDS, but a change in the adal sdk. The core issues from https://github.com/Azure/go-autorest/pull/326 talks about the reconciling the difference between  AAD and ADFS. If we can make NMI response corresponding to the version which is used by application, then we will be able to get it working with old and new adal versions. What do you say ?

lodejard: After taking a look at the api contracts, [they're here](https://github.com/Azure/azure-rest-api-specs/blob/8f2e7ec/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2019-06-01/imds.json#L868), and curling the actual endpoint on a VM with the latest `api-version` (comment above) --- the metadata endpoint does always return strings and never numbers for all of those fields.

I would say it's more correct to always return strings, just like the metadata service does. In that way this PR should work on old and new adal client versions because the NMI response will be identical to what adal is already unmarshalling from the metadata endpoint when NMI is not installed.

lodejard: Here's another way to look at it...

Unmarshal to json.Number works with both wire-forms, so for the client-side like adal it's the best choice for data type.
```
{"example": "42"} -> json.Number=42
{"example": 42}   -> json.Number=42
```

But Marshal from json.Number breaks any client that unmarshals with string, so for the server-side like NMI it causes compat problems
```
json.Number=42 -> {"example": 42} -> json.Number=42
json.Number=42 -> {"example": 42} -> string=Error
```

So even the basic golang example in the ms docs couldn't unmarshal an aad-pod-identity response if it has unquoted numbers.
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-go

aramase: @lodejard Thank you for trying it out. I also went through the code in the sdk and what you are saying makes total sense.

kkmsft:
Thank you @lodejard

lodejard:
Any time @kkmsft, and thank you again for your excellent work on this project. We're very much looking forward to removing service principal secrets from our release pipelines.

Issue #288, opened 2019-07-11
Health probe


kkmsft:

kkmsft:
Released with v1.5 - https://github.com/Azure/aad-pod-identity/releases/tag/v1.5

PR #309, opened 2019-07-25
Liveness probe

kkmsft:
**Reason for Change**:

- Add liveliness at /healthz endpoint for MIC and NMI at 8080 by default.
~~- Add readiness at /ready endpoint for MIC and NMI at 8080 by default.~~
~~- Both liveness and readiness probe return back 200 okay as soon as they are activated from the current run.~~
- The liveliness probe provides "Active" or "Not Active" status as per the state of the component. In case of MIC if the given instance is the elected leader then "Active" string is presented in the /healthz endpoint. For NMI it would show "Active" as soon as the iptables are put in place. 
- Fix the NMI iptables to run instantly instead of waiting for the first ticker instance.

**Issue Fixed**:

#288 

**Notes for Reviewers**:

### DIFF
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -25,7 +25,6 @@ var (
 
 func main() {
 	defer glog.Flush()
-	startTime := time.Now()
 	hostName, err := os.Hostname()
 	if err != nil {
 		glog.Fatalf("Get hostname failure. Error: %+v", err)
@@ -71,9 +70,9 @@ func main() {
 	}
 
 	// Health probe will always report success once its started.
-	// MIC instance will report ready only once its elected the leader
+	// MIC instance will report the contents as "Active" only once its elected the leader
 	// and starts the sync loop.
-	probes.InitAndStart(httpProbePort, startTime, &micClient.SyncLoopStarted)
+	probes.InitAndStart(httpProbePort, &micClient.SyncLoopStarted)
 
 	// Starts the leader election loop
 	micClient.Run()
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -42,7 +42,7 @@ func main() {
 	flag.DurationVar(&leaderElectionCfg.Duration, "leader-election-duration", time.Second*15, "leader election duration")
 
 	//Probe port
-	flag.StringVar(&httpProbePort, "http-probe-port", "8080", "http health and ready probe port")
+	flag.StringVar(&httpProbePort, "http-probe-port", "8080", "http liveliness probe port")
 
 	flag.Parse()
 	if versionInfo {
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/Azure/aad-pod-identity/pkg/mic"
+	"github.com/Azure/aad-pod-identity/pkg/probes"
 	"github.com/Azure/aad-pod-identity/version"
 	"github.com/golang/glog"
 	"k8s.io/client-go/rest"
@@ -19,10 +20,12 @@ var (
 	versionInfo       bool
 	syncRetryDuration time.Duration
 	leaderElectionCfg mic.LeaderElectionConfig
+	httpProbePort     string
 )
 
 func main() {
 	defer glog.Flush()
+	startTime := time.Now()
 	hostName, err := os.Hostname()
 	if err != nil {
 		glog.Fatalf("Get hostname failure. Error: %+v", err)
@@ -38,6 +41,9 @@ func main() {
 	flag.StringVar(&leaderElectionCfg.Namespace, "leader-election-namespace", "default", "namespace to create leader election objects")
 	flag.StringVar(&leaderElectionCfg.Name, "leader-election-name", "aad-pod-identity-mic", "leader election name")
 	flag.DurationVar(&leaderElectionCfg.Duration, "leader-election-duration", time.Second*15, "leader election duration")
+
+	//Probe port
+	flag.StringVar(&httpProbePort, "http-probe-port", "8080", "http health and ready probe port")
 
 	flag.Parse()
 	if versionInfo {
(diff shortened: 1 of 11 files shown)


deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -103,7 +103,7 @@ spec:
         name: iptableslock
       containers:
       - name: nmi
-        image: "mcr.microsoft.com/k8s/aad-pod-identity/nmi:1.5-rc1"
+        image: "testle5.azurecr.io/nmi:le"
```
aramase: Need to switch back to rc1

deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -191,7 +203,7 @@ spec:
       serviceAccountName: aad-pod-id-mic-service-account
       containers:
       - name: mic
-        image: "mcr.microsoft.com/k8s/aad-pod-identity/mic:1.5-rc1"
+        image: "testle5.azurecr.io/mic:le"
```
aramase: Same as above

test/common/k8s/infra/infra.go
```diff
@@ -11,16 +11,22 @@ import (
 )
 
 // CreateInfra will deploy all the infrastructure components (nmi and mic) on a Kubernetes cluster
-func CreateInfra(namespace, registry, nmiVersion, micVersion, templateOutputPath string) error {
-	t, err := template.New("deployment-rbac.yaml").ParseFiles(path.Join("template", "deployment-rbac.yaml"))
+func CreateInfra(namespace, registry, nmiVersion, micVersion, templateOutputPath string, old bool) error {
+	var err error
+	var t *template.Template
+	if !old {
+		t, err = template.New("deployment-rbac.yaml").ParseFiles(path.Join("template", "deployment-rbac.yaml"))
```
aramase: Rather than having 2 different deployment files, we should just template the single file to include new specs/ignore based on a boolean flag.

kkmsft: That sounds good. Will take it up in a future PR.

deploy/infra/master/replicaset/deployment-rbac.yaml
kkmsft: Thanks @aramase

test/e2e/aadpodidentity_test.go
```diff
@@ -102,9 +110,18 @@ var _ = AfterSuite(func() {
...
 var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
+	BeforeEach(func() {
+		testIndex++
+		fmt.Printf("\n\n[%d] %s\n", testIndex, CurrentGinkgoTestDescription().TestText)
+	})
+
 	AfterEach(func() {
 		fmt.Println("\nTearing down the test environment...")
-
+		if CurrentGinkgoTestDescription().Failed {
```
aramase: I don't we should skip cleanup if there is a failure. Instead we should dump the logs here. If we skip clean up, it can cause issues in automation.

test/e2e/aadpodidentity_test.go
```diff
@@ -573,7 +590,7 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 		Expect(ok).To(Equal(true))
 
 		// setup mic and nmi with old releases
-		setupInfra("mcr.microsoft.com/k8s/aad-pod-identity", "1.4", "1.3")
+		setupInfraOld("mcr.microsoft.com/k8s/aad-pod-identity", "1.4", "1.3")
```
aramase: I think instead of 2 functions we should just pass a boolean and use a single template to generate the final deployment.

test/e2e/aadpodidentity_test.go
kkmsft: Yes I agree. This is stop gap before going to the full copy logs and events atleast until that PR is done.

test/e2e/aadpodidentity_test.go
kkmsft: Did not want to make changes in setupinfra all through, instead kept the changes to internal function only.

deploy/infra/master/replicaset/deployment.yaml
```diff
@@ -116,6 +128,18 @@ spec:
           - name: k8s-azure-file
             mountPath: /etc/kubernetes/azure.json
             readOnly: true
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+          initialDelaySeconds: 10
+          periodSeconds: 5
+        readinessProbe:
```
aramase: `ReadinessProbe` is typically used to see if an endpoint is ready to receive traffic. Only when the readinessprobe passes the pod ip is added to list of endpoints in the service that is exposing the pod. We don't have a service with mic or nmi. Is there a reason we are doing readiness probe too?

test/e2e/aadpodidentity_test.go
kkmsft: Ideally we should generate the required template based on version, <= 1.4. When that change is made in future PR, this single instance and setupold function will go away.

cmd/mic/main.go
```diff
@@ -39,6 +41,9 @@ func main() {
 	flag.StringVar(&leaderElectionCfg.Name, "leader-election-name", "aad-pod-identity-mic", "leader election name")
 	flag.DurationVar(&leaderElectionCfg.Duration, "leader-election-duration", time.Second*15, "leader election duration")
 
+	//Probe port
+	flag.StringVar(&httpProbePort, "http-probe-port", "8080", "http health and ready probe port")
```
aramase: nit: `http health and liveness probe port`

pkg/nmi/server/server.go
```diff
@@ -91,6 +103,11 @@ func (s *Server) updateIPTableRules() {
 	ticker := time.NewTicker(time.Second * time.Duration(s.IPTableUpdateTimeIntervalInSeconds))
 	defer ticker.Stop()
 
+	// Run once before the waiting on ticket for the rules to take effect
```
aramase: nit: s/ticket/ticker

kkmsft:
Ran 13 of 13 Specs in 2388.736 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (2388.74s)
PASS
ok      github.com/Azure/aad-pod-identity/test/e2e      2388.765s

Issue #284, opened 2019-07-11
Whitelisting of applications to bypass Aad-pod-identity


kkmsft:

kkmsft:
Closing with release of 1.5 - https://github.com/Azure/aad-pod-identity/releases/tag/v1.5

PR #310, opened 2019-07-25
Whitelisting

aramase:
**Reason for Change**:

There are instances where pods deployed in a cluster need to request token directly instead of using aad-pod-identity. The following changes are done to support that use case - 
- Introduce new CRD - `AzurePodIdentityException`.
- The CRD `spec.PodLabels` field accepts a list of labels. If a pod needs to bypass aad-pod-identity while requesting a token, the label defined in the exception CRD should also be added to the pod. If a match is found, then `NMI` will proxy the request for token without and send the response back without any validation.
- `AzurePodIdentityException` CRD and label matching is done per namespace, which means if the same labels need to be used in multiple namespaces for bypassing pod-identity, a CRD needs to be defined in each namespace required.

**Issue Fixed**:

Feature https://github.com/Azure/aad-pod-identity/issues/284

**Notes for Reviewers**:


aramase:
```bash
Ran 15 of 15 Specs in 2374.242 seconds
SUCCESS! -- 15 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (2374.24s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	2374.271s
```

pkg/apis/aadpodidentity/v1/types.go
```diff
@@ -70,6 +70,18 @@ type AzureAssignedIdentity struct {
 	Status AzureAssignedIdentityStatus `json:"Status"`
 }
 
+//AzurePodIdentityException contains the pod selectors for all pods that don't require
+// NMI to process and request token on their behalf.
+
+//+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
kkmsft: Can you please add a README under the readme directory on how ExceptionList can be used.

test/e2e/template/deployment-rbac.yaml
```diff
@@ -41,6 +41,19 @@ spec:
     plural: azureidentities
   scope: Namespaced
 ---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: azurepodidentityexceptions.aadpodidentity.k8s.io
```
kkmsft: We should add it in the deploy file under master directory as welll. And then move that to the deployment file in the top level folder.

pkg/nmi/server/server.go
```diff
@@ -266,15 +267,15 @@ func (s *Server) isMIC(podNS, rsName string) bool {
 	return false
 }
 
-func (s *Server) getMICToken(logger *log.Entry, rqClientID, rqResource string) ([]byte, int, error) {
+func (s *Server) gotTokenForExceptedPod(logger *log.Entry, rqClientID, rqResource string) ([]byte, int, error) {
```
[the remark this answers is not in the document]
aramase: It's actually only excepted pods. Which other case does it cover?

PR #312, opened 2019-07-26
E2e - collect info on test failure

kkmsft:
**Reason for Change**:

- This PR adds the ability to collect the logs, events, endpoint and pod state, in case of a test failure. 
- Also removed is ginkgo fail fast so that tests can continue if one of them fails.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/aadpodidentity_test.go
+++ b/test/e2e/aadpodidentity_test.go
@@ -116,7 +116,6 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 		if CurrentGinkgoTestDescription().Failed {
 			fmt.Printf("Test failed. Collecting debugging information.")
 			collectDebuggingInfo()
-			return
 		}
 		// Ensure a clean cluster after the end of each test
 		cmd := exec.Command("kubectl", "delete", "AzureIdentity,AzureIdentityBinding", "--all")
@@ -722,7 +721,7 @@ func collectLogs(podName, dir string) {
 
 func collectPods(dir string) {
 	logFile := path.Join(dir, "pods")
-	cmd := exec.Command("bash", "-c", "kubectl get pods > "+logFile)
+	cmd := exec.Command("bash", "-c", "kubectl get pods -o wide > "+logFile)
 	util.PrintCommand(cmd)
 	_, err := cmd.CombinedOutput()
 	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


test/e2e/aadpodidentity_test.go
```diff
@@ -118,8 +114,8 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 	AfterEach(func() {
 		fmt.Println("\nTearing down the test environment...")
 		if CurrentGinkgoTestDescription().Failed {
-			noCleanup = true
-			fmt.Println("Test failed. No cleanup performed")
+			fmt.Printf("Test failed. Collecting debugging information.")
+			collectDebuggingInfo()
```
aramase: If there is an error with the current test, shouldn't we collect the debug info and still continue to do the cleanup? If we return, then we won't cleanup the assigned identities and the identity validator pods. We should get debug info and continue with the rest of `AfterEach`

test/e2e/aadpodidentity_test.go
```diff
@@ -712,6 +708,68 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
...
+	logDirName := path.Join(logsPath, tNow.Format(time.RFC3339))
+	err := os.MkdirAll(logDirName, os.ModePerm)
+	Expect(err).NotTo(HaveOccurred())
+
+	infoFile := path.Join(logDirName, "info")
+	fd, err := os.Create(infoFile)
+	Expect(err).NotTo(HaveOccurred())
+
+	fd.WriteString("Test name: " + CurrentGinkgoTestDescription().TestText + "\n")
+	fd.WriteString("Collecting diagnosics at: " + tNow.Format(time.UnixDate))
```
aramase: nit: s/diagnosics/diagnostics

test/e2e/aadpodidentity_test.go
```diff
@@ -712,6 +708,68 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
...
+		cmd := exec.Command("bash", "-c", "kubectl logs "+p+" > "+logFile)
+		util.PrintCommand(cmd)
+		_, err := cmd.Output()
+		Expect(err).NotTo(HaveOccurred())
+	}
+}
+
+func collectPods(dir string) {
+	logFile := path.Join(dir, "pods")
+	cmd := exec.Command("bash", "-c", "kubectl get pods > "+logFile)
```
aramase: can we do `kubectl get pods -o wide` so we also have the node information for all the pods?

PR #315, opened 2019-07-26
Wrap stats update with lock during parallel sync

aramase:
**Reason for Change**:

Wrap stats update with a lock to prevent concurrent access in the sync loop where nodes are processed in parallel.

**Issue Fixed**:

**Notes for Reviewers**:


pkg/mic/mic.go
```diff
@@ -58,7 +58,9 @@ type Client struct {
 	SyncLoopStarted   bool
 	syncRetryInterval time.Duration
 
-	syncing       int32 // protect against conucrrent sync's
+	syncing    int32 // protect against conucrrent sync's
+	statsMutex sync.Mutex
```
kkmsft: I think this should be a stats property and locking should be internally handled by stats.

aramase: This is really the only place where stats is updated in parallel currently. I think we can switch to moving it to stats and wrapping in structs next.

kkmsft:
Merging with the TODO that we will move the mutex to inside stats before 1.5 release.

Issue #314, opened 2019-07-26
ERROR: logging before flag.Parse in NMI logs


aramase:
**Describe the bug**
This message `ERROR: logging before flag.Parse` shows up in the NMI logs when the probes are initialized and started. This is because glog requires `flag.Parse` to be called before any logging is done. Every time glog is called, it'll contain that error message.

**Steps To Reproduce**

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**
master

**Additional context**

kkmsft:
Closing with #317.

PR #317, opened 2019-07-29
Abstract logging interface for use by multiple components - NMI and MIC

kkmsft:
**Reason for Change**:

**Issue Fixed**:

#314 

**Notes for Reviewers**:


pkg/mic/mic.go
```diff
@@ -848,3 +848,17 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
 	stats.Put(stats.TotalCreateOrUpdate, time.Since(beginAdding))
 	c.statsMutex.Unlock()
 }
+
+// Log - this struct abstracts the logging libraries into single interface.
+type Log struct {
```
aramase: Maybe we should move these into `utils.go` inside mic and nmi? Thoughts?

Issue #313, opened 2019-07-26
Failover causing further azure operations to fail


kkmsft:
**Describe the bug**

When a failover to passive mic happens the following error shows up and then continues to retry the same operations.

0724 02:13:18.473167       1 cloudprovider.go:199] Updating user assigned MSIs on k8s-agentpool1-27200934-2
E0724 02:13:18.641155       1 vm.go:48] compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidRequestContent" Message="The request content was invalid and could not be deserialized: 'Required property 'type' not found in JSON. Path 'identity'.'."
E0724 02:13:18.662425       1 mic.go:734] Getting list of msi's from node k8s-agentpool1-27200934-2 resulted in error identity info is nil

**Steps To Reproduce**
With an identity and pod already assigned, a binding is removed. Right after the MIC removes the identity from the underlying VM, a failover is trigger leading to this error. 
Since e2e did not have copying of logs/info at this point, full set of logs could not be collected, especially the logs from the active MIC pod which was just about getting killed. The above conclusion of the steps to reproduce is arrived after going through various possible scenarios.

**Expected behavior**
The cycle of identity removal should continue as is without any errors.

**AAD Pod Identity version**
latest master

**Kubernetes version**
v1.12.7

**Additional context**

PR #320, opened 2019-07-30
Fix assigned id deletion when no identity exists

aramase:
**Reason for Change**:

When the last underlying identity is deleted from the node, the deletion of assigned identity fails because the resource type is not set during the `CreateOrUpdate` call. This PR ensures when the identity list is nil, the resource type is set to None.

Adds e2e test to valid the scenario.

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/313

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/vm.go
+++ b/pkg/cloudprovider/vm.go
@@ -105,6 +105,10 @@ func (i *vmIdentityInfo) RemoveUserIdentity(id string) error {
 	if i.info.Type == compute.ResourceIdentityTypeNone || i.info.Type == compute.ResourceIdentityTypeSystemAssigned {
 		i.info.IdentityIds = nil
 	}
+	// if the identityids is nil and identity type is not set, then set it to ResourceIdentityTypeNone
+	if i.info.IdentityIds == nil && i.info.Type == "" {
+		i.info.Type = compute.ResourceIdentityTypeNone
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/cloudprovider/vm.go
```diff
@@ -105,6 +105,10 @@ func (i *vmIdentityInfo) RemoveUserIdentity(id string) error {
 	if i.info.Type == compute.ResourceIdentityTypeNone || i.info.Type == compute.ResourceIdentityTypeSystemAssigned {
 		i.info.IdentityIds = nil
 	}
+	// if the identityids is nil and identity type is not set, then set it to ResourceIdentityTypeNone
+	if i.info.IdentityIds == nil && i.info.Type == "" {
```
kkmsft: Thinking more here, shouldn't we just skip the call to cloud overall in this scenario ?

aramase:
```shell
Ran 15 of 16 Specs in 3058.972 seconds
FAIL! -- 15 Passed | 0 Failed | 0 Pending | 1 Skipped
--- FAIL: TestAADPodIdentity (3058.97s)
FAIL
FAIL	github.com/Azure/aad-pod-identity/test/e2e	3058.995s
```

aramase:
```shell
Ran 15 of 16 Specs in 3058.972 seconds
PASS! -- 15 Passed | 0 Failed | 0 Pending | 1 Skipped
--- PASS: TestAADPodIdentity (3058.97s)
PASS
PASS	github.com/Azure/aad-pod-identity/test/e2e	3058.995s
```

PR #322, opened 2019-07-31
Add docs for AzurePodIdentityException

aramase:
**Reason for Change**:

Add instruction for `AzurePodIdentityException`

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,35 @@ If the Azure identity is in the same resource group as your AKS cluster nodes, y
 az role assignment create --role "Managed Identity Operator" --assignee <sp id> --scope <full id of the managed identity>
 ```
 
+### Disable aad-pod-identity for a specific pod/application
+
+NMI pods modify the nodes' iptables to intercept calls to Azure Instance Metadata endpoint. This means any request that's made to the Metadata endpoint will be intercepted by NMI even if the pod doesn't use aad-pod-identity. `AzurePodIdentityException` CRD can be configured to inform aad-pod-identity that any requests to metadata endpoint originating from a pod that matches labels defined in CRD should be proxied without any processing in NMI. NMI will proxy the request to the metdata endpoint and return the token back as is without any validation.
+
+1. Create the `AzurePodIdentityException` with the same label that will be defined in the pod -
+
+```yaml
+apiVersion: "aadpodidentity.k8s.io/v1"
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -188,6 +188,35 @@ If the Azure identity is in the same resource group as your AKS cluster nodes, y
 az role assignment create --role "Managed Identity Operator" --assignee <sp id> --scope <full id of the managed identity>
 ```
 
+### Disable aad-pod-identity for a specific pod/application
+
```
kkmsft: Since the top level readme is growing long, my suggestion is to create a new README under the doc/readmes and then link from table of contents of top README to the new README.

README.md
```diff
@@ -188,6 +188,35 @@ If the Azure identity is in the same resource group as your AKS cluster nodes, y
...
+```yaml
+apiVersion: "aadpodidentity.k8s.io/v1"
+kind: AzurePodIdentityException
+metadata:
+  name: test-exception
+spec:
+  PodLabels:
+    foo: bar
+    app: custom
```
kkmsft: Can you please put a sample template under aad-pod-identity/deploy/demo or some other location in the code so folks can directly use that file.

README.md
```diff
@@ -188,6 +188,35 @@ If the Azure identity is in the same resource group as your AKS cluster nodes, y
...
+``` 
+
+Save this Kubernetes manifest to a file named `aadpodidentityexception.yml`. Replace the PodLabels with a list of desired values and then create the resource on the cluster:
+
+```shell
+kubectl apply -f aadpodidentityexception.yml
+```
+
+2. When creating application pods that will not be using aad-pod-identity for calls to Azure Instance Metadata endpoint, include at least one of the labels in `spec.template.metadata.labels`.
```
kkmsft: Can you please show two examples - on how it will look in the pod specification yaml and other by means of kubectl labels to show how it would look when running.

deploy/infra/master/replicaset/deployment-rbac.yaml
```diff
@@ -206,6 +206,7 @@ spec:
     metadata:
       labels:
         component: mic
+        app: mic
```
kkmsft: Is there some specific reason for adding this label ?

PR #324, opened 2019-08-02
Update base alpine image to 3.10.1

aramase:
**Reason for Change**:

Update base alpine image to `3.10.1`. That is the latest available tag with no vulnerabilities.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ ARG DEMO_VERSION=0.0.0-dev
 ARG IDENTITY_VALIDATOR_VERSION=0.0.0-dev
 RUN make build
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG DEMO_VERSION=0.0.0-dev
 ARG IDENTITY_VALIDATOR_VERSION=0.0.0-dev
 RUN make build
 
-FROM alpine:3.8 AS base
+FROM alpine:3.10.1 AS base
```
kkmsft: Can you please update the busy box image used in the test as well.

PR #325, opened 2019-08-05
Validation and MIC failure e2e test

kkmsft:
**Reason for Change**:

Add a test which runs the identity validation test and the MIC failover parallely.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/common/k8s/pod/pod.go
+++ b/test/common/k8s/pod/pod.go
@@ -165,13 +165,14 @@ func WaitOnDeletion(prefix string) (bool, error) {
 	}
 }
 
-func StopPod(podName string) error {
+// DeletePod - Delete a pod of given name using kubectl
+func DeletePod(podName string) error {
 	cmd := exec.Command("kubectl", "delete", "pod", podName)
 	util.PrintCommand(cmd)
 	_, err := cmd.CombinedOutput()
 
 	if err != nil {
-		errors.Wrap(err, fmt.Sprintf("Pod stop command failed with error: %+v", err))
+		errors.Wrap(err, "Pod delete command failed with error")
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


test/common/k8s/pod/pod.go
```diff
@@ -164,3 +164,14 @@ func WaitOnDeletion(prefix string) (bool, error) {
 		}
 	}
 }
+
+func StopPod(podName string) error {
```
aramase: Can we rename this to `DeletePod` as that's what it's really doing?

test/common/k8s/pod/pod.go
```diff
@@ -164,3 +164,14 @@ func WaitOnDeletion(prefix string) (bool, error) {
...
+	cmd := exec.Command("kubectl", "delete", "pod", podName)
+	util.PrintCommand(cmd)
+	_, err := cmd.CombinedOutput()
+
+	if err != nil {
+		errors.Wrap(err, fmt.Sprintf("Pod stop command failed with error: %+v", err))
```
aramase: Since the error is wrapped we don't need to fmt

test/e2e/aadpodidentity_test.go
```diff
@@ -873,8 +874,108 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
...
+		}(iterations)
+
+		wg.Wait()
+		fmt.Printf("Done with running validator test and disrupting MIC")
+	})
 })
 
+func runValidatorTest(iterations int) {
+	defer GinkgoRecover()
+	fmt.Println("Starting identity validator. Iteration: 0")
```
aramase: We can remove this log as it's printed in line `915`

Issue #332, opened 2019-08-08
'size of request headers too long error' from nmi


smira:
**Describe the bug**

Seems to happen randomly when pods get launched - it fails to acquire MSI identity.

`mic` logs look clean - identity is assigned, but `nmi` log shows:

```
time="2019-08-08T16:15:16Z" level=info msg="matched identityType:0 clientid:62e24516-4c4a-4540-950e-c4e07d6218a8 resource:https://management.core.windows.net/" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=10.15.192.25
time="2019-08-08T16:15:16Z" level=error msg="failed to get service principal token for pod:woke/dev-westus-cron-woke-run-1565280900-7zjft, adal: Refresh request failed. Status Code = '400'. Response body: <!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 4.01//EN\"\"http://www.w3.org/TR/html4/strict.dtd\">\r\n<HTML><HEAD><TITLE>Bad Request</TITLE>\r\n<META HTTP-EQUIV=\"Content-Type\" Content=\"text/html; charset=us-ascii\"></HEAD>\r\n<BODY><h2>Bad Request - Request Too Long</h2>\r\n<hr><p>HTTP Error 400. The size of the request headers is too long.</p>\r\n</BODY></HTML>\r\n" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=10.15.192.25
time="2019-08-08T16:15:16Z" level=info msg="Status (403) took 67315333 ns" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=10.15.192.25
```

Seems similar to #323 

**Steps To Reproduce**

Not a high load, single pod is launched every 10-15 minutes. Pod sleeps 10 seconds on startup, but still this error.

**Expected behavior**

Identities is assigned as expected.

**AAD Pod Identity version**

1.5

**Kubernetes version**

1.13.5

**Additional context**

aramase:
@smira Thank you for trying out aad-pod-identity. I have a few questions for you as I'm unable to recreate this - 
1. Which client library is the pod `dev-westus-cron-woke-run-1565280900-7zjft` using for requesting token? Which version of the client library is it using?
2. Did this work fine with the old release `mic:1.3 and nmi:1.4`? 
3. Does restarting the nmi help resolve the issue? It did in case of #323. So curious if this is consistent.

[...]

PR #333, opened 2019-08-08
Append NMI version to the `User-Agent` for adal only once

smira:
Before the fix, append happened each time
`GetServicePrincipalTokenFromMSIWithUserAssignedID` was called, which
was blowing up HTTP header size (or header value size).

This also updates adal UserAgent for any call from this library, not
only for the one mentioned above.

**Reason for Change**:

Fixes a bug, makes UA header consistent.

**Issue Fixed**:

Fixes: #332

(not tested yet)

**Notes for Reviewers**:


aramase:
Thank you for this fix @smira

smira:
thanks for the quick turnaround and new release :+1:

PR #334, opened 2019-08-08
Change 'updateStrategy' for nmi DaemonSet to `RollingUpdate`

smira:
Without explicit strategy set, it defaults to `OnDelete` which means
`nmi` pods will be updated only when they are terminated, which doesn't
happen often obviously and it's quite unpredictable.

**Reason for Change**:

With this change, `kubectl apply` correctly upgrades and restarts all
the components when new verison is applied.

**Issue Fixed**:

none

**Notes for Reviewers**:


kkmsft:
Thank you, @smira

PR #343, opened 2019-08-14
Redact client id in NMI logs

aramase:
**Reason for Change**:

Redact client id in NMI logs

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/nmi/server/server_test.go
+++ b/pkg/nmi/server/server_test.go
@@ -31,7 +31,7 @@ func TestGetTokenForMatchingIDBySP(t *testing.T) {
 		Spec: aadpodid.AzureIdentitySpec{
 			Type:           aadpodid.ServicePrincipal,
 			TenantID:       "tid",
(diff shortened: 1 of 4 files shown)


pkg/nmi/server/server_test.go
```diff
@@ -31,7 +31,7 @@ func TestGetTokenForMatchingIDBySP(t *testing.T) {
 		Spec: aadpodid.AzureIdentitySpec{
 			Type:           aadpodid.ServicePrincipal,
 			TenantID:       "tid",
-			ClientID:       "cid",
+			ClientID:       "aabc0000-a83v-9h4m-000j-2c0a66b0c1f9",
```
kkmsft: can the test assert the finally generated string ?

aramase: We can't check the log to assert, but I think I can move it to a util function and unit test that.

PR #345, opened 2019-08-16
Fix list identities at scale

aramase:
**Reason for Change**:

Getting list of assigned identities fails when the number of assigned identities is > 500. This is to fix issue with pods at scale.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -10,6 +10,7 @@ import (
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
 	"github.com/Azure/aad-pod-identity/pkg/stats"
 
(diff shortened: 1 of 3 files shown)


pkg/crd/crd.go
```diff
@@ -262,11 +263,19 @@ func (c *Client) ListBindings() (res *[]aadpodid.AzureIdentityBinding, err error
 // ListAssignedIDs returns a list of azureassignedidentities
 func (c *Client) ListAssignedIDs() (res *[]aadpodid.AzureAssignedIdentity, err error) {
 	begin := time.Now()
+
 	ret, err := c.AssignedIDListWatch.List(v1.ListOptions{})
 	if err != nil {
 		glog.Error(err)
-		return nil, err
+		return
```
kkmsft: return nil, err ?

PR #347, opened 2019-08-19
Re-enable the UTs by using fake rest client

### LINKED ISSUE #243: Reenable UTs calling mock ListWatcher
**Describe the bug**
The UTs which call ListWatcher are commented out in #240  because of the bug: https://github.com/kubernetes/client-go/issues/352. This issue is raised to track, so that we can reenable when we find a solution for this issue.

**Steps To Reproduce**

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**

**Additional context**

kkmsft:
**Reason for Change**:

When fake rest client is used, the UT eventually hits the issue in https://github.com/kubernetes/client-go/issues/352. This PR uses the rest client from the fake package and also adds the corresponding HTTP clients to return back the pods to simulate the original package behavior.

**Issue Fixed**:

Fixes #243 

**Notes for Reviewers**:

### DIFF
--- a/pkg/k8s/client_test.go
+++ b/pkg/k8s/client_test.go
@@ -1,12 +1,25 @@
 package k8s
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
 	"testing"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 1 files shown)


pkg/k8s/client_test.go
```diff
@@ -32,79 +45,166 @@ func TestGetSecret(t *testing.T) {
...
+	kubeClient := &KubeClient{ClientSet: client, PodListWatch: podListWatch}
 
 	time.AfterFunc(time.Duration(1200*time.Millisecond), func() {
-		fakeClient.CoreV1().Pods("default").Create(pod)
+		testClientSet.AddPod(testPodName, testPodNs, testPodIP)
 	})
 
 	start := time.Now()
-	podNs, podName, err := kubeClient.GetPodName(podIP)
+	podNs, podName, _, _, err := kubeClient.GetPodInfo(podIP)
```
aramase: Should we also check the replica set name and labels are returned correctly?

kkmsft: Will add in future PR.

PR #348, opened 2019-08-19
Readme fixes for namespaced instructions

kkmsft:
**Reason for Change**:

This PR reorganizes the namespaced pods instructions to point to a README of its own.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/docs/readmes/README.namespaced.md
+++ b/docs/readmes/README.namespaced.md
@@ -1,0 +1,40 @@
+# Match Pods in the Namespace
+> Available from (1.3.0-mic-1.4.0-nmi release)[https://github.com/Azure/aad-pod-identity/releases/tag/1.3.0-mic-1.4.0-nmi]
+
+By default, AAD Pod Identity matches pods to identities across namespaces. To match only pods in the namespace containing `AzureIdentity`, use one of these techniques:
+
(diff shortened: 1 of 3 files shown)


docs/readmes/README.namespaced.md
```diff
@@ -0,0 +1,40 @@
...
+      name: <a-idname>
+      annotations:
+        aadpodidentity.k8s.io/Behavior: namespaced
+    spec:
+      type: 0
+      ResourceID: /subscriptions/<subid>/resourcegroups/<resourcegroup>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>
+      ClientID: <clientId>
+    ```
+
+* Add the `--forceNamespaced` command line argument or set the `FORCENAMESPACED=true` environment variable when starting both the MIC and NMI components.
```
aramase: We should say doing this will enable forced namespace for all identity/bindings/pod created. Also, I think we should mention even though the force namespace directive is used, the azure assigned identity is always created in default namespace.

aramase: > Also, I think we should mention even though the force namespace directive is used, the azure assigned identity is always created in default namespace.

Please ignore this.

PR #351, opened 2019-08-20
README now includes in-pod token retrieval URL

MateMalice:
**Reason for Change**:

The README.md contains the endpoint URL from which to retrieve a token from a node, but not the URL for retrieving tokens from within pods themselves. I think that specifying this URL in the main README would help people figure out how to use this package more easily, so I have updated the relevant section in order to specify the endpoint URL that should be called to retrieve an OAuth2.0 token from within a pod. I've tested the URL specified and it worked on my AKS deployment.

**Issue Fixed**:

**Notes for Reviewers**:


aramase:
thank you @MateMalice!

MateMalice:
No problems - hoping this extra info helps others get this working! :-)

Issue #366, opened 2019-08-28
Delete assigned identities when node no longer found


aramase:
**Describe the bug**
Delete assigned identities for a node, if the node is no longer bound to the cluster.

**Steps To Reproduce**

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**

**Additional context**

PR #367, opened 2019-08-29
Clean up assigned identities if node not found

aramase:
**Reason for Change**:

Delete assigned identities for the node, when node no longer exists.

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/366

**Notes for Reviewers**:
Pending testing + running e2e

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -9,9 +9,10 @@ import (
 	"github.com/golang/glog"
 
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	"github.com/Azure/aad-pod-identity/pkg/logger"
+	inlog "github.com/Azure/aad-pod-identity/pkg/logger"
 	"github.com/Azure/aad-pod-identity/pkg/stats"
 
-	log "github.com/Azure/aad-pod-identity/pkg/logger"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -33,8 +34,8 @@ type Client struct {
 
 // ClientInt ...
 type ClientInt interface {
-	Start(exit <-chan struct{})
-	SyncCache(exit <-chan struct{}, initial bool)
+	Start(exit <-chan struct{}, log inlog.Logger)
+	SyncCache(exit <-chan struct{}, initial bool, log inlog.Logger)
 	RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureAssignedIdentity) error
 	CreateAssignedIdentity(assignedIdentity *aadpodid.AzureAssignedIdentity) error
 	UpdateAzureAssignedIdentityStatus(assignedIdentity *aadpodid.AzureAssignedIdentity, status string) error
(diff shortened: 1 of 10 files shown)


pkg/crd/crd.go
```diff
@@ -244,13 +244,17 @@ func (c *Client) Start(exit <-chan struct{}) {
...
+	c.SyncCache(exit, true)
 	glog.Info("CRD informers started")
 }
 
 // SyncCache synchronizes cache
-func (c *Client) SyncCache(exit <-chan struct{}) {
+func (c *Client) SyncCache(exit <-chan struct{}, initial bool) {
 	if !cache.WaitForCacheSync(exit) {
+		if !initial {
+			glog.Error("Cache could not be synchronized")
```
kkmsft: StartLite can get called from NMI, which will result in glog call instead of logrus call. We could use the log interface (like in the probe) to use the right logging calls.

pkg/mic/mic.go
```diff
@@ -237,6 +238,11 @@ func (c *Client) Sync(exit <-chan struct{}) {
 		begin := time.Now()
 		workDone := false
 
+		cacheTime := time.Now()
+		// sync cache to get the latest data
```
kkmsft: Please elaborate the comment here something which says the following - "since we could have done creates in the previous cycle and which has not yet been propagated we might end up doing the cycles again. Syncing the cache avoids such situations. "

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
 
 	node, err := c.NodeClient.Get(nodeName)
 	if err != nil {
-		glog.Errorf("Lookup of identities on node %s resulted in error %v", nodeName, err)
+		glog.Warningf("Unable to get node %s while updating user msis. Error %v", nodeName, err)
+
+		if !strings.Contains(err.Error(), "not found") {
+			return
```
kkmsft: Looks like in the current code as well as in the new addition we are just swallowing the errors which come from here. Shouldn't we propagate this up ?

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+		}
+
+		// node is no longer found in the cluster, all the assigned identities that were created in this sync loop
+		// and those that already exist for this node need to be deleted.
+		for _, createID := range nodeTrackList.assignedIDsToCreate {
+			binding := createID.Spec.AzureBindingRef
+
+			err = c.removeAssignedIdentity(&createID)
+			if err != nil {
+				c.EventRecorder.Event(binding, corev1.EventTypeWarning, "binding remove error",
```
kkmsft: assigned identity removal ?

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+					fmt.Sprintf("Removing assigned identity binding %s node %s for pod %s resulted in error %v", binding.Name, createID.Spec.NodeName, createID.Name, err.Error()))
+				glog.Error(err)
+				continue
+			}
+			c.EventRecorder.Event(binding, corev1.EventTypeNormal, "binding removed",
```
kkmsft: assigned id removal

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+				fmt.Sprintf("Binding %s removed from node %s for pod %s", binding.Name, createID.Spec.NodeName, createID.Spec.Pod))
+		}
+
+		for _, deleteID := range nodeTrackList.assignedIDsToDelete {
+			binding := deleteID.Spec.AzureBindingRef
+
+			err = c.removeAssignedIdentity(&deleteID)
+			if err != nil {
+				c.EventRecorder.Event(binding, corev1.EventTypeWarning, "binding remove error",
```
kkmsft: binding ==> assigned id

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
 	node, err := c.NodeClient.Get(nodeName)
 	if err != nil {
-		glog.Errorf("Lookup of identities on node %s resulted in error %v", nodeName, err)
+		glog.Warningf("Unable to get node %s while updating user msis. Error %v", nodeName, err)
+
+		if !strings.Contains(err.Error(), "not found") {
+			return
+		}
+
+		// node is no longer found in the cluster, all the assigned identities that were created in this sync loop
```
kkmsft: Can we move this to a utility function and then add UTs.

pkg/mic/mic.go
```diff
@@ -730,7 +736,41 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
 
 	node, err := c.NodeClient.Get(nodeName)
 	if err != nil {
-		glog.Errorf("Lookup of identities on node %s resulted in error %v", nodeName, err)
+		glog.Warningf("Unable to get node %s while updating user msis. Error %v", nodeName, err)
+
+		if !strings.Contains(err.Error(), "not found") {
```
kkmsft: Please log if the error is - not found.

pkg/mic/mic.go
aramase: I've added a unit test in MIC to test this specific scenario.

pkg/mic/mic.go
```diff
@@ -185,7 +186,7 @@ func (c *Client) Start(exit <-chan struct{}) {
 
 	wg.Add(1)
 	go func() {
-		c.CRDClient.Start(exit)
+		c.CRDClient.Start(exit, Log{})
```
kkmsft: I was thinking we should embed a struct in miclient and use that here and other places..

aramase: ok, I'll take a look.

Issue #361, opened 2019-08-26
Update identities on VMSS only once


aramase:
**Describe the bug**

**Steps To Reproduce**

**Expected behavior**

**AAD Pod Identity version**

**Kubernetes version**

**Additional context**

PR #379, opened 2019-09-11
update identity in vmss only once

aramase:
**Reason for Change**:

Update identity in vmss only once as identity is across all nodes in vmss and not just single node.

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/361

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -18,7 +18,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/golang/glog"
 	yaml "gopkg.in/yaml.v2"
-	corev1 "k8s.io/api/core/v1"
 )
 
 // Client is a cloud provider client
@@ -31,10 +30,10 @@ type Client struct {
 
 // ClientInt client interface
 type ClientInt interface {
-	RemoveUserMSI(userAssignedMSIID string, node *corev1.Node) error
-	AssignUserMSI(userAssignedMSIID string, node *corev1.Node) error
-	UpdateUserMSI(addUserAssignedMSIIDs []string, removeUserAssignedMSIIDs []string, node *corev1.Node) error
-	GetUserMSIs(node *corev1.Node) ([]string, error)
+	RemoveUserMSI(userAssignedMSIID, name string, isvmss bool) error
+	AssignUserMSI(userAssignedMSIID, name string, isvmss bool) error
+	UpdateUserMSI(addUserAssignedMSIIDs, removeUserAssignedMSIIDs []string, name string, isvmss bool) error
+	GetUserMSIs(name string, isvmss bool) ([]string, error)
 }
 
 // NewCloudProvider returns a azure cloud provider client
@@ -153,8 +152,8 @@ func withInspection() autorest.PrepareDecorator {
 }
 
 // GetUserMSIs will return a list of all identities on the node
-func (c *Client) GetUserMSIs(node *corev1.Node) ([]string, error) {
-	idH, _, err := c.getIdentityResource(node)
+func (c *Client) GetUserMSIs(name string, isvmss bool) ([]string, error) {
+	idH, _, err := c.getIdentityResource(name, isvmss)
 	if err != nil {
 		glog.Errorf("GetUserMSIs: get identity resource failed with error %v", err)
 		return nil, err
@@ -168,8 +167,8 @@ func (c *Client) GetUserMSIs(node *corev1.Node) ([]string, error) {
 }
 
 // UpdateUserMSI will batch process the removal and addition of ids
-func (c *Client) UpdateUserMSI(addUserAssignedMSIIDs []string, removeUserAssignedMSIIDs []string, node *corev1.Node) error {
-	idH, updateFunc, err := c.getIdentityResource(node)
+func (c *Client) UpdateUserMSI(addUserAssignedMSIIDs, removeUserAssignedMSIIDs []string, name string, isvmss bool) error {
+	idH, updateFunc, err := c.getIdentityResource(name, isvmss)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


pkg/mic/vmss.go
```diff
@@ -96,6 +96,11 @@ func makeVMSSID(r azure.Resource) string {
 	return path.Join(r.SubscriptionID, r.ResourceGroup, r.ResourceName)
 }
 
+func getVMSSName(vmssID string) string {
+	_, resourceName := path.Split(vmssID)
```
kkmsft: Is it possible to have an erroneous vmssID, if so how do we handle this ?

pkg/mic/mic.go
```diff
@@ -874,31 +865,72 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
 
-	for _, deleteID := range nodeTrackList.assignedIDsToDelete {
-		binding := deleteID.Spec.AzureBindingRef
+func (c *Client) consolidateVMSSNodes(nodeMap map[string]trackUserAssignedMSIIds) {
+	vmssMap := make(map[string][]string)
 
-		err := c.removeAssignedIdentity(&deleteID)
+	for nodeName, nodeTrackList := range nodeMap {
+		node, err := c.NodeClient.Get(nodeName)
+		if err != nil && !strings.Contains(err.Error(), "not found") {
```
kkmsft: const ?

pkg/mic/mic.go
```diff
@@ -874,31 +865,72 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+			glog.Errorf("Unable to get node %s. Error %v", nodeName, err)
+			continue
+		}
+		if err != nil && strings.Contains(err.Error(), "not found") {
+			glog.Warningf("Unable to get node %s while updating user msis. Error %v", nodeName, err)
```
kkmsft: spurious line ?

pkg/mic/mic.go
```diff
@@ -874,31 +865,72 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+
+			// node is no longer found in the cluster, all the assigned identities that were created in this sync loop
+			// and those that already exist for this node need to be deleted.
+			go c.cleanUpAllAssignedIdentitiesOnNode(nodeName, nodeTrackList)
```
kkmsft: Is there a reason for us to perform this async ?

kkmsft: Is there anything preventing another sync loop to come and perform the same operations again ?

pkg/mic/mic.go
```diff
@@ -874,31 +865,72 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
 		}
 		c.EventRecorder.Event(binding, corev1.EventTypeNormal, "binding removed",
-			fmt.Sprintf("Binding %s removed from node %s for pod %s", binding.Name, createID.Spec.NodeName, createID.Spec.Pod))
+			fmt.Sprintf("Binding %s removed from node %s for pod %s", binding.Name, deleteID.Spec.NodeName, deleteID.Spec.Pod))
 	}
+}
 
-	for _, deleteID := range nodeTrackList.assignedIDsToDelete {
-		binding := deleteID.Spec.AzureBindingRef
+func (c *Client) consolidateVMSSNodes(nodeMap map[string]trackUserAssignedMSIIds) {
```
kkmsft: Can you please add comments here on overall approach of this function.

pkg/mic/vmss.go
aramase: If the vmssID is erroneous, then we fail while trying to create the client and not perform any operations.

pkg/mic/mic.go
aramase: Used only here, so didn't define as a const.

pkg/mic/mic.go
aramase: We don't want the clean up to block the node processing for rest of the nodes/identities. Good point about preventing sync loop doing again. We can create a WaitGroup at sync loop and pass that for consolidating (used if there is a clean up for node not existing) and for update node identities. That way we can block until all node processing is complete.

pkg/mic/mic.go
```diff
@@ -866,39 +857,82 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+		for i := 0; i < len(vmssNodes); i++ {
+			vmssTrackList.addUserAssignedMSIIDs = append(vmssTrackList.addUserAssignedMSIIDs, nodeMap[vmssNodes[i]].addUserAssignedMSIIDs...)
+			vmssTrackList.removeUserAssignedMSIIDs = append(vmssTrackList.removeUserAssignedMSIIDs, nodeMap[vmssNodes[i]].removeUserAssignedMSIIDs...)
+			vmssTrackList.assignedIDsToCreate = append(vmssTrackList.assignedIDsToCreate, nodeMap[vmssNodes[i]].assignedIDsToCreate...)
+			vmssTrackList.assignedIDsToDelete = append(vmssTrackList.assignedIDsToDelete, nodeMap[vmssNodes[i]].assignedIDsToDelete...)
+			vmssTrackList.isvmss = true
+
+			delete(nodeMap, vmssNodes[i])
+
+			nodeMap[getVMSSName(vmssName)] = vmssTrackList
```
kkmsft: Can you please add comments which specifies that we are using nodeMap with both kind of keys.

pkg/mic/vmss.go
kkmsft: Can we please handle the error for the path.Split returning an empty string ?

pkg/mic/mic.go
```diff
@@ -866,39 +857,82 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+
+	// aggregate vmss nodes into vmss name
+	for vmssName, vmssNodes := range vmssMap {
+		if len(vmssNodes) < 1 {
+			continue
+		}
+
+		vmssTrackList := trackUserAssignedMSIIds{}
+
+		for i := 0; i < len(vmssNodes); i++ {
```
kkmsft: Why index rather than range usage ?

pkg/mic/mic.go
```diff
@@ -866,39 +857,82 @@ func (c *Client) updateUserMSI(newAssignedIDs []aadpodid.AzureAssignedIdentity,
...
+		if err != nil && !strings.Contains(err.Error(), "not found") {
+			glog.Errorf("Unable to get node %s. Error %v", nodeName, err)
+			continue
+		}
+		if err != nil && strings.Contains(err.Error(), "not found") {
+			glog.Warningf("Unable to get node %s while updating user msis. Error %v", nodeName, err)
+			wg.Add(1)
+			// node is no longer found in the cluster, all the assigned identities that were created in this sync loop
+			// and those that already exist for this node need to be deleted.
+			go c.cleanUpAllAssignedIdentitiesOnNode(nodeName, nodeTrackList, wg)
```
kkmsft: If we use anonymous function, we don't have to pass wg to cleanUpAllAssignedIdentitiesOnNode. Is that idiomatic way to use ?

pkg/mic/mic.go
```diff
@@ -310,8 +310,15 @@ func (c *Client) Sync(exit <-chan struct{}) {
 			c.getListOfIdsToAssign(*addList, nodeMap)
 		}
 
+		var wg sync.WaitGroup
```
kkmsft: Can you please add a comment near nodeMap make to indicate that the key can be now vmss or vm name.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -153,8 +152,8 @@ func withInspection() autorest.PrepareDecorator {
 }
 
 // GetUserMSIs will return a list of all identities on the node
```
kkmsft: nit:s/Node/Node or VMSS depending on the value of isVMSS

pkg/mic/mic.go
aramase: I'm passing the same wg for consolidating vmss node (which perform the cleanup if reqd) and update node deps. That way all our processing is done async.

PR #396, opened 2019-09-23
Add --block-instance-metadata flag

### LINKED ISSUE #377: Block pod access to metadata API, except where explicitly required
**Describe the request**
In multi-tenant Kubernetes clusters, access to the metadata service should be blocked to prevent tenants from accessing sensitive data, such as customData. This is particularly necessary for the `/instance` endpoint, but more generically, this request is to only allow access to what is needed, and not the entire metadata service.

The intent of AAD Pod Identity is to prevent pods from having access greater than what it should -- the principle of least privilege. Without blocking access to the rest of the metadata API, pods can get more information than is desired, especially in cases of the pod being compromised.

As AAD Pod Identity is a proxy (NMI), it should block access to the rest of the metadata API, outside of the necessary `/identity` endpoints (or other endpoints as seen fit). To accommodate anyone who does not want this feature, it should be enabled/disabled by a feature flag.

Network Policies can block this access on a per-namespace basis, excepting where Identities are needed or where the namespace admins require access to manage network policies. In all, network policies are very useful to address this problem, but only in a subset of contexts.

dharmab:
**Reason for Change**:

This PR adds a `--block-instance-metadata` flag to NMI that blocks the `/metadata/instance` path in the metadata API with HTTP 403. This is an optional security feature to prevent a Pod from viewing potentially sensitive data in the metadata API, such as VM customData or OS version.

**Issue Fixed**:

Fixes #377 

**Notes for Reviewers**:

I'm a noob in this repo- is there a standard integration test for this project? Or do I need to test this manually by deploying it into a cluster?


pkg/nmi/server/server.go
```diff
@@ -74,7 +76,11 @@ func (s *Server) Run() error {
 	mux.Handle("/metadata/identity/oauth2/token/", appHandler(s.msiHandler))
 	mux.Handle("/host/token", appHandler(s.hostHandler))
 	mux.Handle("/host/token/", appHandler(s.hostHandler))
-	mux.Handle("/", appHandler(s.defaultPathHandler))
+	if s.BlockInstanceMetadata {
+		mux.Handle("/", http.NotFoundHandler())
```
kkmsft: Can we please change this to StatusForbidden to reflect that we are blocking the calls ?

kkmsft:
Thank you, @dharmab for the patch. There are UTs which can be run on demand from GitHub. @aramase has initiated the UT run. There are e2e tests which can be run by following the instructions here: https://github.com/Azure/aad-pod-identity/tree/master/test/e2e. But in this case since the regular code path is not effected , please try manual testing and let us know. We have regular running e2e internally (based on the instructions above, but not yet exposed via Azure pipelines) - between @aramase  and me we will get the patch through the e2e and merge it.

pkg/nmi/server/server.go
dharmab: I was torn between 403 and 404 here. In the end I picked 404 since it makes the code smaller since [the http package provides a handler for 404](https://golang.org/pkg/net/http/#NotFoundHandler). 404 is also sometimes used instead of 403 deliberately; for example, GitHub returns a 404 if you try to navigate to a private repo that you don't have access to.

Thoughts on the tradeoff between slightly more code and the slightly more correct HTTP code?

dharmab:
Thanks @kkmsft. How could I contribute a test for this code path? Also, do any docs need to be updated?

pkg/nmi/server/server.go
dharmab: Hah, turns out the NotFoundHandler implementation is trivial:

```go
// NotFoundHandler returns a simple request handler
// that replies to each request with a ``404 page not found'' reply.
func NotFoundHandler() Handler { return HandlerFunc(NotFound) }
```

So this can easily be changed to 403 😄

dharmab:
Updated:

- HTTP 404 -> HTTP 403
- Only block /metadata/instance to avoid blocking legitimate useful paths like /scheduledevents

dharmab:
@aramase @kkmsft - Could this PR get an updated review? Thanks!

pkg/nmi/server/server.go
```diff
@@ -458,6 +463,11 @@ func (s *Server) defaultPathHandler(logger *log.Entry, w http.ResponseWriter, r
 	w.Write(body)
 }
 
+// forbiddenHandler responses to any request with HTTP 403 Forbidden
+func forbiddenHandler(w http.ResponseWriter, r *http.Request) {
+	http.Error(w, "", http.StatusForbidden)
```
aramase: since NMI is the one blocking the requests when deployed, I'm thinking maybe we should have an error msg along with 403. 

@kkmsft thoughts?

dharmab: That's probably a good idea, in case someone gets confused and tries to contact Azure support.

dharmab:
@aramase @kkmsft this is ready for another review 😄

dharmab:
Rebase to resolve conflicting flags

kkmsft:
Thank you, @dharmab. Will look at the new changes today.

kkmsft:
> Thanks @kkmsft. How could I contribute a test for this code path? Also, do any docs need to be updated?

With respect to the e2e, here is the link to how to run e2e:
https://github.com/Azure/aad-pod-identity/tree/master/test/e2e
There are tests present in the following location:
https://github.com/Azure/aad-pod-identity/blob/master/test/e2e/aadpodidentity_test.go
You can follow the examples there to have a test which uses environment which specifically enabled this flag and deploys the image and runs tests to validate the functionality. Please let us know if you run into any issues with this.

Here is the documentation location:
https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.md#features
Here we have features added specifically, so you can add a feature here and specific readme for this feature.

kkmsft:
@dharmab - can you please let us know if you have plans to add e2e and doc in the same PR ?

dharmab:
@kkmsft I'm planning to update the doc in this PR and follow up with tests in a second PR

dharmab:
Docs added. Ready for merge. Thanks!

dharmab:
Another rebase to resolve conflicts

kkmsft:
Thank you @dharmab

PR #398, opened 2019-09-28
Update client go

kkmsft:
**Reason for Change**:

- Update client-go to avoid thread leak.
   - The older go-client had thread leak which would lead to high cpu usage. Fixes by updating the client-go to 1.15.3 related client version. The following has details about the thread leak https://github.com/kubernetes/kubernetes/issues/71277.
   - The newer client-go has been made case sensitive because of which we need to have our fields to be case senstivie. Hence the changes in the types.go to update the json tags. The following issue has more details - https://github.com/kubernetes/kubernetes/issues/64612.
  -  There has been changes in the usage for leader election. MIC code had to be changed to work with this.

- Simple command to test listing and accept GOOS from env
- Fix cache sync calls
- Misc. logging changes

**Issue Fixed**:

**Notes for Reviewers**:
TODO 
- Fix the usage of exit channel
- Run e2e to validate
- Add e2e to validate backward compatibility with older 1.5 and 1.3-1.4


kkmsft:
Summarizing 1 Failure:

[Fail] Kubernetes cluster using aad-pod-identity [It] should assign identity with init containers 
/Users/kkmsft/projects/aad-pod-identity/go/src/github.com/Azure/aad-pod-identity/test/e2e/aadpodidentity_test.go:1216

Ran 17 of 19 Specs in 3898.244 seconds
FAIL! -- 16 Passed | 1 Failed | 0 Pending | 2 Skipped
--- FAIL: TestAADPodIdentity (3898.24s)
FAIL
FAIL    github.com/Azure/aad-pod-identity/test/e2e      3898.289s
make: *** [e2e] Error 1

cmd/simple/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+	flag.Set("logtostderr", "true")
+	flag.Set("v", "10")
+
+	flag.Parse()
+
+	if versionInfo {
+		version.PrintVersionAndExit()
+	}
+	klog.Infof("Starting mic process. Version: %v. Build date: %v", version.MICVersion, version.BuildDate)
```
aramase: Since we are keeping this code in the master branch, can we update the logs to say `Starting simple process` and remove the mic related logs?

pkg/apis/aadpodidentity/v1/types.go
```diff
@@ -166,48 +166,48 @@ const (
...
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	PodLabels         map[string]string `json:"podLabels"`
+	PodLabels         map[string]string `json:"PodLabels"`
 }
 
 // AzurePodIdentityExceptionStatus ...
 type AzurePodIdentityExceptionStatus struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Status            string `json:"status"`
+	Status            string `json:"Status"`
```
aramase: why are we changing this status field to `Status`? Isn't it similar to what other CRDs have defined?

cmd/simple/main.go
aramase: Also, do we want to keep this in the master branch?

pkg/mic/mic.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/aad-pod-identity/version"
 	"golang.org/x/sync/semaphore"
 	corev1 "k8s.io/api/core/v1"
+	kubeErrors "k8s.io/apimachinery/pkg/api/errors"
```
aramase: can we rename this to `apierrors` instead of `kubeErrors`?

pkg/mic/mic.go
```diff
@@ -41,6 +42,18 @@ type NodeGetter interface {
 	Start(<-chan struct{})
 }
 
+// TypeUpgradeConfig - configuration aspects of type related changes required for client-go upgrade.
+type TypeUpgradeConfig struct {
+	CMTypeUpgradeKey  string
```
aramase: Can we add a comment here what this key signifies?

pkg/mic/mic.go
```diff
@@ -205,10 +245,64 @@ func (c *Client) NewLeaderElector(clientSet *kubernetes.Clientset, recorder reco
...
+			}
+			klog.Infof("Type upgrade completed !!")
+			// Upgrade completed so update the data with the upgrade key.
+			if cm.Data == nil {
+				cm.Data = make(map[string]string)
+			}
+			cm.Data[c.TypeUpgradeCfg.CMTypeUpgradeKey] = version.MICVersion
+			_, err = c.CMClient.Update(cm)
+			if err != nil {
+				return fmt.Errorf("type upgrade annotation update on %s failed. error: %+v", c.TypeUpgradeCfg.CMTypeUpgradeKey, err)
```
aramase: `updating config map key for %s failed`

pkg/mic/mic.go
```diff
@@ -205,10 +245,64 @@ func (c *Client) NewLeaderElector(clientSet *kubernetes.Clientset, recorder reco
...
+			}
+		} else {
+			klog.Infof("Type upgrade status configmap found from version: %s. Proceeding without type upgrade !", v)
```
aramase: nit: Skipping type upgrade

pkg/mic/mic.go
```diff
@@ -205,10 +245,64 @@ func (c *Client) NewLeaderElector(clientSet *kubernetes.Clientset, recorder reco
...
+	}
+	return nil
+}
+
 // Start ...
 func (c *Client) Start(exit <-chan struct{}) {
 	klog.V(6).Infof("MIC client starting..")
 
+	if err := c.UpgradeTypeIfRequired(); err != nil {
+		klog.Fatalf("Upgrade failed with error: %v", err)
```
aramase: I think we should say type upgrade failed with error. I think we need to make this error message descriptive so the user what upgrade failed here

pkg/mic/mic.go
```diff
@@ -275,6 +369,13 @@ func (c *Client) Sync(exit <-chan struct{}) {
 		begin := time.Now()
 		workDone := false
 
+		cacheTime := time.Now()
+
+		// There is a delay in data propogation to cache. It's possible that the creates performed in the previous sync cycle
+		// are not propogated before this sync cycle began. In order to avoid redoing the cycle, we sync cache again.
+		c.CRDClient.SyncCacheAll(exit, false)
```
aramase: Were you able to confirm that this cache sync with the go client update has no leaks?

pkg/mic/mic.go
```diff
@@ -417,19 +522,22 @@ func (c *Client) createDesiredAssignedIdentityList(
 	newAssignedIDs := make(map[string]aadpodid.AzureAssignedIdentity)
 
 	for _, pod := range listPods {
+		klog.V(6).Infof("checking pod: %s", pod.Name)
```
aramase: can we also add the pod namespace here to the log?

pkg/mic/mic.go
```diff
@@ -417,19 +522,22 @@ func (c *Client) createDesiredAssignedIdentityList(
...
 		if pod.Spec.NodeName == "" {
 			//Node is not yet allocated. In that case skip the pod
 			klog.V(2).Infof("Pod %s/%s has no assigned node yet. it will be ignored", pod.Namespace, pod.Name)
 			continue
 		}
 		crdPodLabelVal := pod.Labels[aadpodid.CRDLabelKey]
+		klog.V(6).Infof("Label value: %v", crdPodLabelVal)
```
aramase: I think we should also add the podns and podname in this log

pkg/mic/mic.go
```diff
@@ -417,19 +522,22 @@ func (c *Client) createDesiredAssignedIdentityList(
...
 		if crdPodLabelVal == "" {
 			//No binding mentioned in the label. Just continue to the next pod
 			klog.V(2).Infof("Pod %s/%s has correct %s label but with no value. it will be ignored", pod.Namespace, pod.Name, aadpodid.CRDLabelKey)
 			continue
 		}
 		var matchedBindings []aadpodid.AzureIdentityBinding
 		for _, allBinding := range *listBindings {
+			klog.V(6).Infof("Check the binding: %s", allBinding.Spec.Selector)
```
aramase: same as above

deploy/demo/aadpodidentity.yaml
```diff
@@ -4,5 +4,5 @@ metadata:
  name: demo-aad1
 spec:
  type: 0
- ResourceID: RESOURCE_ID
- ClientID: CLIENT_ID 
+ resourceID: RESOURCE_ID
```
aramase: We should update these manifests right before we cut the release as this will not work with current release.

deploy/demo/aadpodidentitybinding.yaml
```diff
@@ -3,5 +3,5 @@ kind: AzureIdentityBinding
 metadata:
   name: demo-azure-id-binding
 spec: 
-  AzureIdentity: "demo-aad1"
-  Selector: "demo"
\ No newline at end of file
+  azureIdentity: "demo-aad1"
+  selector: "demo"
```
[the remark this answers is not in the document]
aramase: ditto as above

Issue #402, opened 2019-10-01
mic crashes when VMSS managed identity is not authorized


ritazh:
**Describe the bug**
When using pod identity in a cluster running VMSS with system assigned identity, mic pod crashes when the system assigned identity does not have the right permission to perform role assignment.

**Steps To Reproduce**
- Create a k8s v1.14 cluster with aks-engine with `"useManagedIdentity": true`
- Ensure the cluster is running keyvault flexvol in kube-system and image used is `mcr.microsoft.com/k8s/flexvolume/keyvault-flexvolume:v0.0.14`
- Install pod identity: `kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml`, currently mic is at 1.5.2.
- Create azureidentity
```yaml
apiVersion: "aadpodidentity.k8s.io/v1"
kind: AzureIdentity
metadata:
 name: ritademo
spec:
 type: 0
 ResourceID: /subscriptions/xxx/resourceGroups/xxx/providers/Microsoft.ManagedIdentity/userAssignedIdentities/ritademo
 ClientID: xxx
```
- Create azureidentitybinding

```yaml
apiVersion: "aadpodidentity.k8s.io/v1"
kind: AzureIdentityBinding
metadata:
 name: demo1-azure-identity-binding
spec:
 AzureIdentity: ritademo
 Selector: demo
```
- Schedule pod deployment with keyvault flexvol in the cluster:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-flex-kv-podid
  labels:
    aadpodidbinding: demo
spec:
  containers:
  - name: nginx-flex-kv-podid
    image: nginx
    volumeMounts:
    - name: test
      mountPath: /kvmnt
      readOnly: true
  volumes:
  - name: test
    flexVolume:
      driver: "azure/kv"
      options:
        usepodidentity: "true" 
        keyvaultname: ritakv
        keyvaultobjectname: secret
        keyvaultobjecttype: secret 
        tenantid: xxx
```

**Expected behavior**
Pod creation should fail. MIC should log the error, but it should not be crashing.

**AAD Pod Identity version**
1.5.2

**Kubernetes version**
1.14.6

**Additional context**

Here is the log dump:

[...]

PR #403, opened 2019-10-01
Check identity ids exists before return

aramase:
**Reason for Change**:

Check identity ids exists.

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/402

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/vm.go
+++ b/pkg/cloudprovider/vm.go
@@ -124,5 +124,8 @@ func (i *vmIdentityInfo) AppendUserIdentity(id string) bool {
 }
 
 func (i *vmIdentityInfo) GetUserIdentityList() []string {
+	if i.info.IdentityIds == nil {
+		return []string{}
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vm.go
```diff
@@ -124,5 +124,8 @@ func (i *vmIdentityInfo) AppendUserIdentity(id string) bool {
 }
 
 func (i *vmIdentityInfo) GetUserIdentityList() []string {
+	if i.info.IdentityIds == nil {
```
kkmsft: Can we also check if i.Info is nil as well before accessing IdentityIds.

aramase: We already check that in here - https://github.com/Azure/aad-pod-identity/blob/master/pkg/cloudprovider/cloudprovider.go#L163-L165

PR #406, opened 2019-10-02
Scale test

aramase:
**Reason for Change**:

- Use map for sync loop computation

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -273,7 +273,7 @@ func (c *Client) Sync(exit <-chan struct{}) {
 			continue
 		}
 
-		currentAssignedIDs, err := c.CRDClient.ListAssignedIDs()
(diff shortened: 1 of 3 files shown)


pkg/mic/mic.go
```diff
@@ -500,91 +503,77 @@ func (c *Client) matchAssignedID(x *aadpodid.AzureAssignedIdentity, y *aadpodid.
...
+	for assignedIDName, oldAssignedID := range old {
+		newAssignedID, exists := new[assignedIDName]
+		idMatch := false
+		if exists {
+			idMatch = c.matchAssignedID(oldAssignedID, newAssignedID)
+		}
+		// assigned identity exists in the desired list too which means
+		// it should not be deleted
+		if exists && idMatch {
+			continue
```
kkmsft: Do we want to simplify ?
if exists {
        if c.match... {
          continue
        }
}

PR #424, opened 2019-10-16
Add application/json header for all return paths

kkmsft:
**Reason for Change**:

**Issue Fixed**:

**Notes for Reviewers**:


pkg/nmi/server/server.go
```diff
@@ -152,6 +152,7 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		"req.path":   r.URL.Path,
 		"req.remote": parseRemoteAddr(r.RemoteAddr),
 	})
+	w.Header().Set("Content-Type", "application/json")
```
aramase: Now that we are doing this in `ServeHTTP` can we get rid of the duplicate from here - https://github.com/Azure/aad-pod-identity/blob/master/pkg/nmi/server/server.go#L462

PR #431, opened 2019-10-23
Adding support for whitelisting of user-defined managed identities

dmeytin:
**Reason for Change**:
In some cases a user-defined managed identity could be removed from the AAD

**Issue Fixed**:
This PR is adding whitelist of user defined managed identities what will never be removed from AAD

**Notes for Reviewers**:
The change is as following:
- mic main  accepts new cmd parameter *whiteListedIdentitiesParam* - comma separated list of identities that shouldn't be deleted from AAD
- at *updateUserMSI* function we compare *removeUserAssignedMSIIDs* and *whiteListedIdentities* and only missing values are removed from AAD
- helm chart and Readme are changed accordingly

This feature is extremely important for migrating existing production clusters to aad-pod-identity.
In particular, all our existing deployments will use user-defined identity as one per team.
Reuse of the same identity to multiple pods caused the identity to be deleted from AAD when it still in use. 
This change will solve the issue.

Best regards,
Dmitry

PR #441, opened 2019-11-21
fix identity removal

aramase:
**Reason for Change**:

With the change for whitelisting identities - https://github.com/Azure/aad-pod-identity/pull/431, non-immutable identities are no longer being cleaned up as expected. 

**Issue Fixed**:

This PR fixes the identity removal from non-immutable identities and also performs the check for immutable/non-immutable right at the start.

**Notes for Reviewers**:

### DIFF
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -103,7 +103,10 @@ func main() {
 	config.Burst = int(clientQPS)
 	glog.Infof("Client QPS set to: %v. Burst to: %v", config.QPS, config.Burst)
 
-	immutableUserMSIsList := strings.Split(immutableUserMSIs, ",")
+	var immutableUserMSIsList []string
+	if immutableUserMSIs != "" {
+		immutableUserMSIsList = strings.Split(immutableUserMSIs, ",")
+	}
 	glog.Infof("immutable identities are %v", immutableUserMSIsList)
 
 	micClient, err := mic.NewMICClient(cloudconfig, config, forceNamespaced, syncRetryDuration, &leaderElectionCfg, enableScaleFeatures, createDeleteBatch, immutableUserMSIsList)
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -120,7 +120,8 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
 
 	var immutableUserMSIsMap map[string]bool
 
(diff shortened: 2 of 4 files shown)


pkg/mic/mic.go
```diff
@@ -120,7 +120,7 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
 
 	var immutableUserMSIsMap map[string]bool
 
-	if immutableUserMSIsList != nil {
+	if immutableUserMSIsList != nil && len(immutableUserMSIsList) > 0 {
```
kkmsft: Isn't len check enough since that checks for nil ?

cmd/mic/main.go
```diff
@@ -103,7 +103,10 @@ func main() {
 	config.Burst = int(clientQPS)
 	glog.Infof("Client QPS set to: %v. Burst to: %v", config.QPS, config.Burst)
 
-	immutableUserMSIsList := strings.Split(immutableUserMSIs, ",")
+	var immutableUserMSIsList []string
+	if immutableUserMSIs != "" {
```
kkmsft: Can we please rename this to ImmutableUserMSIsParam or rename the ImmutableUserMSIs in the client to ImmutableUserMSIsMap ?

test/e2e/aadpodidentity_test.go
```diff
@@ -386,7 +386,7 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 
 		// Delete pod identity to verify that the VM identity did not get deleted
 		waitForDeployDeletion(identityValidator)
-		cmd := exec.Command("kubectl", "delete", "AzureIdentity,AzureIdentityBinding,AzureAssignedIdentity", "--all")
```
kkmsft: Is it possible to add this as separate commit so that we can track back why we changed this ?

cmd/mic/main.go
aramase: It is already `ImmutableUserMSIsMap` in the client.

pkg/mic/mic.go
aramase: Done.

test/e2e/aadpodidentity_test.go
aramase: Done.

cmd/mic/main.go
aramase: Done!

PR #442, opened 2019-11-22
Add check in e2e test to confirm removal of userMSI

kkmsft:
**Reason for Change**:

Bugs in deletion logic of the underlying identity is only captured in the test which checks for pre existing identities. This PR adds this check to the delete test so that we can find it more early.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/aadpodidentity_test.go
+++ b/test/e2e/aadpodidentity_test.go
@@ -194,11 +194,32 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
 	It("should delete the AzureAssignedIdentity if the deployment is deleted", func() {
 		setUpIdentityAndDeployment(keyvaultIdentity, "", "1")
 
+		podName, err := pod.GetNameByPrefix(identityValidator)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(podName).NotTo(Equal(""))
+
+		// Get the name of the node to assign the identity to
+		nodeName, err := pod.GetNodeName(podName)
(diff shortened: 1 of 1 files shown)


test/e2e/aadpodidentity_test.go
```diff
@@ -194,11 +194,32 @@ var _ = Describe("Kubernetes cluster using aad-pod-identity", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 		Expect(ok).To(Equal(true))
+
+		resourceManager, err := getResourceManager(n)
+		Expect(err).NotTo(HaveOccurred())
+
+		userAssignedIdentities, err := resourceManager.GetUserAssignedIdentities()
+		Expect(err).NotTo(HaveOccurred())
+		// Check if the user assigned Identity has been removed from the underlying VM/VMSS.
+		Expect(len(userAssignedIdentities)).To(Equal(0))
```
aramase: Instead of checking the length, can we just check the `keyvaultIdentity` is not in the map?

aramase: Adding a note here based on our discussion offline - get the list of identities on the node before the test is run. After the test is run, get the list of identities again from the node and check the list is same as what it was at the start of the test.

- Helps validate the identity that was assigned as part of the test has been successfully removed.
- Ensures no new identities were assigned by MIC

aramase:
Thank you for the PR. Closing this as there have been changes to the test suite since this and it's no longer required.

PR #451, opened 2019-12-09
Update metrics name and refactor

aramase:
**Reason for Change**:

**Issue Fixed**:

**Notes for Reviewers**:


go.mod
```diff
@@ -27,14 +26,12 @@ require (
...
 	github.com/pkg/errors v0.8.0
-	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829 // indirect
 	github.com/sirupsen/logrus v1.2.0
 	github.com/spf13/pflag v1.0.1
 	go.opencensus.io v0.22.0
 	golang.org/x/sync v0.0.0-20190227155943-e225da77a7e6
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
-	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
-	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
+	google.golang.org/appengine v1.4.0
```
kkmsft: Any ideas on how app engine is getting called ?

aramase: Running `go mod why` shows it's not in use by main module. So now it removed it.

docs/readmes/README.monitoring.md
```diff
@@ -33,30 +33,38 @@ Histogram that tracks the duration (in seconds) it takes for Assigned identity d
...
-Histogram that tracks the duration (in seconds) it takes for cloud provider operations.Broken down by operation type.
+Histogram that tracks the duration (in seconds) it takes for cloud provider operations. Broken down by operation type.
 
 **11. aadpodidentity_kubernetes_api_operations_errors_count**
 
-Counter that tracks the cumulative number of kubernetes api operations errors.Broken down by operation type.
\ No newline at end of file
+Counter that tracks the cumulative number of kubernetes api operations errors. Broken down by operation type.
+
+**12. aadpodidentity_imds_operations_errors_count**
```
kkmsft: don't we need one which count the IMDS operations without error ?

aramase: The user can get that metrics by filtering # of requests with response code. I don't think we need a metric for the # of calls we make.

pkg/metrics/metrics.go
```diff
@@ -252,3 +297,28 @@ func RegisterAndExport(port string, log log.Logger) error {
 	log.Infof("Registered and exported metrics on port %s", port)
 	return nil
 }
+
+// ReportIMDSOperationError reports IMDS error count
+func (r *Reporter) ReportIMDSOperationError(operation string) error {
```
kkmsft: Quick question - why is this preferable vs two methods - one which counts and one which measure latency ? and then separate flags which can be used in a switch case to perform different operations like - ImdsOperationsErrorCount

pkg/auth/auth.go
```diff
@@ -10,95 +10,105 @@ import (
...
 // GetServicePrincipalTokenFromMSI return the token for the assigned user
 func GetServicePrincipalTokenFromMSI(resource string) (*adal.Token, error) {
 	begin := time.Now()
+	var err error
+
+	defer func() {
+		if err != nil {
+			reporter.ReportIMDSOperationError(metrics.AdalTokenFromMSIOperationName)
+			return
+		}
```
kkmsft: Shouldn't we count the operations themselves or is the idea to look at the latency reports and find the total number based on the key value of count in the latency metrics.

pkg/cloudprovider/vmss.go
```diff
@@ -63,23 +58,29 @@ func NewVMSSClient(config config.AzureConfig, spt *adal.ServicePrincipalToken) (
 func (c *VMSSClient) CreateOrUpdate(rg string, vmssName string, vm compute.VirtualMachineScaleSet) error {
 	// Set the read-only property of extension to null.
 	//vm.Resources = nil
-
 	ctx := context.Background()
 	begin := time.Now()
+	var err error
+
+	defer func() {
+		if err != nil {
```
kkmsft: Do we want to check if the reporter is initialized (nil check and error logging if not) here and in other such blocks ?

pkg/crd/crd.go
```diff
@@ -315,46 +315,54 @@ func (c *Client) syncCache(exit <-chan struct{}, initial bool, cacheSyncs ...cac
...
-	c.reporter.Report(
-		metrics.AssignedIdentityDeletionCountM.M(1),
-		metrics.AssignedIdentityDeletionDurationM.M(metrics.SinceInSeconds(begin)))
+	defer func() {
+		if err != nil {
+			c.reporter.ReportKubernetesAPIOperationError(metrics.AssignedIdentityDeletionOperationName)
+			return
+		}
+		c.reporter.Report(
+			metrics.AssignedIdentityDeletionCountM.M(1),
```
kkmsft: Here we are differencing in the pattern - we are counting the deletes instead of looking at the number from the latency counters. What do you say ?

pkg/metrics/metrics.go
```diff
@@ -51,51 +77,59 @@ var (
 		"Total number of assigned identity deletion operations",
 		stats.UnitDimensionless)
 
-	// NodeManagedIdentityOperationsDurationM is a measure that tracks the duration in seconds of nodemanagedidentity operations.
-	NodeManagedIdentityOperationsDurationM = stats.Float64(
-		nodeManagedIdentityOperationsDurationName,
-		"Duration in seconds of node managed identity operations",
+	// NMIOperationsDurationM is a measure that tracks the duration in seconds of nmi operations.
```
kkmsft: nmi ==> NMI

pkg/metrics/metrics.go
```diff
@@ -51,51 +77,59 @@ var (
...
+	NMIOperationsDurationM = stats.Float64(
+		nmiOperationsDurationName,
+		"Duration in seconds for nmi operations",
```
kkmsft: nit: nmi ==> NMI

pkg/nmi/server/server.go
```diff
@@ -189,27 +189,34 @@ func (fn appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			namespace,
+			ns,
 			resource,
-			metrics.NodeManagedIdentityOperationsDurationM.M(metrics.SinceInSeconds(start)))
+			metrics.NMIOperationsDurationM.M(metrics.SinceInSeconds(start)))
 	}
 }
 
-func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.Request) {
+func (s *Server) hostHandler(logger *log.Entry, w http.ResponseWriter, r *http.Request) (ns string) {
```
kkmsft: Are we returning namespaces from this ? Also, does not look like all return paths return the namespace ?

PR #457, opened 2019-12-13
Support multiple identities in Helm chart

### LINKED ISSUE #428: Support creation of multiple AzureIdentity and AzureIdentityBinding resources in Helm chart
The current Helm chart only supports creating a single `AzureIdentity` and `AzureIdentityBinding`. In real world, different pods will likely have different identities with isolated permissions.

While customising the current chart implementation is possible, would be nice if this change came from upstream, making it easier to update the chart to newer versions without a bunch of manual YAML editing.

amirschw:
**Reason for Change**:
Support creation of multiple AzureIdentity and AzureIdentityBinding resources in Helm chart.

**Issue Fixed**:
Closes #428.

**Notes for Reviewers**:
Again wasn't sure about the chart version. Decided to update it since this is a breaking change to chart users.


amirschw:
Thanks you, @aramase. That's right, users will have to change their values file when upgrading. I have added a short explanation of the required change in the chart's `README.md` file.

charts/aad-pod-identity/templates/identities.yaml
```diff
@@ -1,24 +1,26 @@
...
+  - apiVersion: "aadpodidentity.k8s.io/v1"
+    kind: AzureIdentity
+    metadata:
+      name: {{ .name }}
+      namespace: {{ default $.Release.Namespace .namespace }}
+      labels:
+        {{- include "aad-pod-identity.labels" $ | nindent 8 }}
+    spec:
+      type: {{ .type }}
+      ResourceID: {{ required "resourceID is required!" .resourceID }}
```
aramase, line 14: The client-go version has been updated to latest in 1.6 and this means the CRD fields are now case sensitive. More details on the changes required - https://github.com/Azure/aad-pod-identity#v160-breaking-change.

Can you update the fields to use the right case when you get a chance?

aramase:
@amirschw Could you also rebase the PR when you get a chance?

charts/aad-pod-identity/templates/identities.yaml
amirschw: Done

charts/aad-pod-identity/templates/identities.yaml
```diff
@@ -1,24 +1,26 @@
...
-kind: AzureIdentityBinding
-metadata:
-  name: {{ .Values.azureIdentityBinding.name }}
-  namespace: {{ template "aad-pod-identity.azureidentitybinding.namespace" . }}
-  labels:
-    {{- include "aad-pod-identity.labels" . | nindent 4 }}
-spec:
-  azureIdentity: {{ .Values.azureIdentity.name }}
-  selector: {{ required ".Values.azureIdentityBinding.selector is required!" .Values.azureIdentityBinding.selector }}
+apiVersion: v1
```
chewong, line 1: We should add a length check for `.Values.azureIdentities` in case the user didn't supply any identities.

aramase: Good catch.

aristosvo: Is the relation between binding and identity always 1:1? Shouldn’t we separate it in two lists in the templates?

chewong: I think 1:1 makes sense. Each `AzureIdentity` should only have one `AzureIdentityBinding`, which can be consumed by multiple pods.

aristosvo: 1:1 makes sense in _most_ scenarios, but there must be a reason to have two separate CRDs in the first place. Which legitimates the question whether we should define them as 1 in the configuration and templates. I don’t mind going forward with the current implementation, but I think it’s a valid question in itself.

What are your thoughts, @aramase ?

amirschw: If I remember correctly from testing it, we'll just have an empty list of identities if the user didn't supply any. Nonetheless, let me add this validation.

aramase: @aristosvo Keeping it 1:1 makes sense. It makes it easier for users to configure multiple ids and bindings all at once.

PR #459, opened 2019-12-16
Create internal api for aadpodidentity

cnadolny:
**Reason for Change**:
Changes:
- Add internal API, change all project references to operate on internal API
- Add conversion functions and tests
  - I initially tried to use generated functions created by conversion-gen, but this project has dependencies on older versions of apimachinery and client-go. It's not possible to easily update those dependencies either, as it would require a refactor of LeaderElection to use contexts rather than channels.
- Refactored parts of V1 api and moved pieces to internal (channel types, util.go function)

Conversions:
- Conversions occur from V1 -> Internal inside of the List Binding/IDs/Exception functions, when we are grabbing the list of resources from the informer
- Conversion from Internal -> V1 only occurs when creating an AssignedIdentity, when the HTTP POST call is made.

**Issue Fixed**:
Part 1 for fixing #237
Will create another CR for the V2 API changes

**Notes for Reviewers**:
Unit tests pass, manually tested that it works, e2e passed


pkg/apis/aadpodidentity/types.go
```diff
@@ -0,0 +1,228 @@
+package aadpodidentity
+
+import (
+	api "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
```
kkmsft: Is 'internal' directory within the aadpodidentity more appropriate for these files ?

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,124 @@
...
+			Pod:              assignedIdentity.Spec.Pod,
+			PodNamespace:     assignedIdentity.Spec.PodNamespace,
+			NodeName:         assignedIdentity.Spec.NodeName,
+			Replicas:         assignedIdentity.Spec.Replicas,
+		},
+		Status: aadpodidv1.AzureAssignedIdentityStatus(assignedIdentity.Status),
+	}
+}
+
+func ConvertV1AzurePodIdentityExceptionToInternalAzurePodIdentityException(idException aadpodidv1.AzurePodIdentityException) (residException aadpodid.AzurePodIdentityException) {
```
kkmsft: Can we please move to earlier in the file, since most of the v1->internal conversions are present there.

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,124 @@
...
+	return aadpodid.AzurePodIdentityException{
+		TypeMeta:   idException.TypeMeta,
+		ObjectMeta: idException.ObjectMeta,
+		Spec: aadpodid.AzurePodIdentityExceptionSpec{
+			ObjectMeta: idException.Spec.ObjectMeta,
+			PodLabels:  idException.Spec.PodLabels,
+		},
+		Status: aadpodid.AzurePodIdentityExceptionStatus(idException.Status),
```
kkmsft: Don't we need an internal Exception to v1 exception ? Is it because we don't create the exception list programatically ? Could you please add a comment to indicate that ?

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,124 @@
+package conversion
```
kkmsft: Would it be better to have version conversions move to their respective directories ? Wouldn't this file grow too long when we add more versions ?

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,124 @@
...
+			NodeName:         assignedIdentity.Spec.NodeName,
+			Replicas:         assignedIdentity.Spec.Replicas,
+		},
+		Status: aadpodid.AzureAssignedIdentityStatus(assignedIdentity.Status),
+	}
+}
+
+func ConvertInternalBindingToV1Binding(identityBinding aadpodid.AzureIdentityBinding) (resIdentityBinding aadpodidv1.AzureIdentityBinding) {
+	return aadpodidv1.AzureIdentityBinding{
+		TypeMeta:   identityBinding.TypeMeta,
```
kkmsft: where are we changing updating the apiVersion to v1 ?

kkmsft: SetGroupVersionKind ?

test/common/k8s/azurepodidentityexception/azurepodidentityexception.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os/exec"
 	"path"
 
-	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
```
kkmsft: Isn't this reflecting user scenario and thus have v1 being used ?

test/e2e/aadpodidentity_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"sync"
 	"time"
 
-	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
```
kkmsft: Why is internal version used for user tests ?

test/common/k8s/azureidentitybinding/azureidentitybinding.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os/exec"
 	"path"
 
-	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
+	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
```
kkmsft: Same here: why is internal version used for user scenarios as against the v1 ?

pkg/mic/mic.go
```diff
@@ -118,7 +119,7 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
 
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: clientSet.CoreV1().Events("")})
-	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: aadpodid.CRDGroup})
+	recorder := eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: aadpodv1.CRDGroup})
```
kkmsft: I think we should use recorderv1 to use for cases where the v1 is used to distinguish the code. Thoughts ?

pkg/apis/aadpodidentity/types.go
cnadolny: golang has restrictions on how directories with name internal are used, left as is

pkg/conversion/conversion.go
cnadolny: Updated version via SetGroupVersionKind

pkg/conversion/conversion.go
cnadolny: Yes - we only grab the list of exceptions and operate on them internally, but they are not created by the project, which is why I did not create the internal -> v1 function. Will add a comment

pkg/mic/mic.go
cnadolny: Updated this to just use the internal API, as the CRDGroup is just a static string.

kkmsft:
Thank you, @cnadolny ! 
Merging this since we have root causes issue with the test failures and also some other PRs need rebasing based on this merge. 
Please paste the e2e snapshot when you get a chance.

cnadolny:
I had to debug one more E2E test, but it turned out to be a nonissue with the code and they did all pass!

Tearing down the test suite environment...
$ kubectl delete -f template/_output/default-deployment.yaml --now --ignore-not-found
$ kubectl delete -f template/busyboxds.yaml
$ kubectl delete deploy --all
$ kubectl get nodes -ojson
# Removing identity 'immutable-identity' from aks-nodepool1-30527383-vmss...
# Removing system assigned identity from aks-nodepool1-30527383-vmss...
# Removing system assigned identity from aks-nodepool1-30527383-vmss...
# Removing system assigned identity from aks-nodepool1-30527383-vmss...

Ran 20 of 20 Specs in 4355.987 seconds
SUCCESS! -- 20 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAADPodIdentity (4355.99s)
PASS
ok  	github.com/Azure/aad-pod-identity/test/e2e	4356.010s
MININT-H5157QD:aad-pod-identity administrador$

PR #496, opened 2020-02-03
Add pod informer to watch pod created and list all existing pods 

bowu24:
**Reason for Change**:

Add support for windows AAD pod identity.

**Issue Fixed**:

Add pod informer to watch pod created and list all existing pods 

**Notes for Reviewers**:


bowu24:
This is the idea implementation for the design doc we send. This is  just draft of PR and just for initial review purpose.

0xbliang:
@bowu24 I think you added me by mistake. But I can't unassigned myself.

pkg/mic/mic.go
```diff
@@ -113,7 +113,8 @@ func NewMICClient(cloudconfig string, config *rest.Config, isNamespaced bool, sy
 	}
 	klog.V(1).Infof("CRD client initialized")
 
-	podClient := pod.NewPodClient(informer, eventCh)
```
bhaskardeep: Do you need any change in the MIC. The MIC would continue to run in Linux, and the NMI would run in Linux and windows.

pkg/nmi/server/linuxServer.go
```diff
@@ -0,0 +1,83 @@
+// +build linuxos
+
```
bhaskardeep: The server can just be one too, right?

pkg/nmi/server/server.go
```diff
@@ -513,24 +491,6 @@ func copyHeader(dst, src http.Header) {
 	}
 }
 
-func handleTermination() {
```
bhaskardeep: Whats is the windows equivalent of this?

pkg/nmi/server/windowsServer.go
```diff
@@ -0,0 +1,83 @@
...
+type WindowsServer struct {
+	Server *Server
+}
+
+// Run runs the specified Server.
+func (s *WindowsServer) Run() error {
+
+	var wg sync.WaitGroup
+
+	wg.Add(1)
```
bhaskardeep: This will fire one thread right, so do you still need thread synchronization using waitgroup?

pkg/nmi/server/windowsServer.go
```diff
@@ -0,0 +1,83 @@
...
+
+			fmt.Printf("Pod Node Name and Pod IP: %s %s \n", pod.Spec.NodeName, pod.Status.PodIP)
+			if s.Server.NodeName == pod.Spec.NodeName {
+				applyRoutePolicy(pod.Status.PodIP)
+			}
+		}
+	}
+}
+
+func (s *WindowsServer) ApplyExistingPods() {
```
bhaskardeep: nit rename: ApplyRoutePolicyForExistingPods()

bowu24: Updated

pkg/mic/mic.go
bowu24: Ideally, there is no change in MIC. However, MIC also use the podclient. We modify the current podclient code pod.NewPodClient to add a new pod channel, which will be used in NMI code.  so we need to make the corresponding change in MIC. The change will never change any function and logic for MIC.

pkg/nmi/server/linuxServer.go
bowu24: For server, I think we need to files, one is for linux syscall, and the other is for windows syscall. If we combine to one file, build will not work.

pkg/nmi/server/windowsServer.go
bowu24: I think you are right. No need to have sync.WaitGroup here. Updated the code.

pkg/nmi/server/server.go
bowu24: Yes. I am adding the code to handle pod delete event.

bhaskardeep: My question is, should we need to, since I believe endpoint would be deleted after pod delete. Let me ask this to the folks in the thread.

bhaskardeep: My question is, should we need to, since I believe endpoint would be deleted after pod delete. Let me ask this to the folks in the thread.

bowu24: Yeah. There are two cases here: 
1. When pod is deleted, it will fire a pod deletion event. If the endpoint would be deleted after pod deleted, then it will be easy for us to do nothing, here, right? Otherwise, our code needs to call DeleteRoute through proxy agent. It is a little tricky here since once the pod is deleted, we may not be able to get the podip in pod delete event, which mean we may need to cache this info during pod creation event. During pod deletion, we can always get the pod name, which will help us to filter the corresponding pod ip if we cache both info during pod creation event.

2. When the nmi pod is deleted or stopped for some reason, should we list all the existing pods and call DeleteRoute through proxy agent.

bowu24: For now, I added the change for no 2.

cmd/nmi/main.go
```diff
@@ -114,7 +114,7 @@ func main() {
 	if err = metrics.RegisterAndExport(*prometheusPort); err != nil {
 		klog.Fatalf("Could not register and export metrics: %+v", err)
 	}
-	if err := s.Run(); err != nil {
-		klog.Fatalf("%s", err)
-	}
+
+	// Run nmi server
+	server.RunServer(s)
```
bhaskardeep: >RunServer [](start = 8, length = 9)

Is there any reason we removed the error handling here?

pkg/nmi/server/server.go
```diff
@@ -70,13 +88,13 @@ func NewServer(micNamespace string, blockInstanceMetadata bool) *Server {
 		MICNamespace:          micNamespace,
 		BlockInstanceMetadata: blockInstanceMetadata,
 		Reporter:              reporter,
+		PodClient:             podClient,
+		PodObjChannel:         podObjCh,
 	}
 }
 
 // Run runs the specified Server.
 func (s *Server) Run() error {
```
bhaskardeep: >Run [](start = 17, length = 3)

Nit: Can't this method be called by the nmi/main.go, which in turn runs the linux or windows method to handle routing. We can name those as routing_windows.go and routing_linux.go, and the method could be named as ConfigureRouting. Having two server is kind of confusing now. I feel that way we have a clear line of calls, i.e. the client (main.go) call the server run(), and the server run() configures routing and then listens to the endpoints.

pkg/nmi/server/linuxServer.go
```diff
@@ -0,0 +1,90 @@
...
+	iptables "github.com/Azure/aad-pod-identity/pkg/nmi/iptables"
+	"k8s.io/klog"
+)
+
+type LinuxServer struct {
+	Server *Server
+}
+
+func RunServer(s *Server) {
```
alukyan: >RunServer [](start = 5, length = 9)

Why do you have to duplicate "Run" logic for both Windows and Linux.
I would simply keep original server.Run() implementation and add one parameter to the Run() function

pkg/pod/pod.go
```diff
@@ -98,6 +105,52 @@ func (c *Client) GetPods() (pods []*v1.Pod, err error) {
 	return listPods, nil
 }
 
+// ListPods returns list of all pods
+func (c *Client) ListPods() (pods []*v1.Pod, err error) {
```
alukyan: >ListPods [](start = 17, length = 8)

I am concerned about performance impact. When MIC singleton makes a query that lists pods is one thing. But NMI is running on every node so the impact of such query is multiplied with the number of nodes in the cluster. Is there a way to limit the query to get pods on a specific node?

pkg/pod/pod.go
```diff
@@ -98,6 +105,52 @@ func (c *Client) GetPods() (pods []*v1.Pod, err error) {
...
+	if !exists || err != nil {
+		err := fmt.Errorf("Could not get Pod %s", obj.(*v1.Pod))
+		klog.Error(err)
+		return nil, err
+	}
+
+	pod = currentPod.(*v1.Pod)
+	for pod.Status.PodIP == "" || pod.Spec.NodeName == "" {
+		fmt.Printf("Sleep 1 second to wait for pod ip and node name\n")
+		time.Sleep(1 * time.Second)
```
alukyan: Seeing such sleeps in the code is usually an indicator of some timing issue in the code. Why do you need to sleep for 1 second?

pkg/mic/mic.go
alukyan: I agree. We should not change MIC code. Is there any way to avoid this change?

---
In reply to: [374486614](https://github.com/Azure/aad-pod-identity/pull/496#discussion_r374486614) [](ancestors = 374486614)

pkg/pod/pod.go
bowu24: When AKS starts to deploy a pod, the pod creation event will be fired. It always happen before the pod is completed created, which means when we detect the pod creation event, the pod IP is not assigned to the pod yet. So add this check here. I tested a bunch of pods and most of them needs wait 4~5 seconds to get the pod id assigned.

cmd/nmi/main.go
bowu24: No, it is not removed. Instead, it is moved inside server.RunServer(s)  function.

pkg/nmi/server/linuxServer.go
bowu24: Good points. I will check the way you suggested.

pkg/nmi/server/server.go
bowu24: both 1 and 2 are added.

pkg/nmi/server/server.go
bowu24: Andrey also gives very similar comments like this. I will check if I can implement this as you suggested.

pkg/nmi/server/windowsServer.go
```diff
@@ -0,0 +1,145 @@
...
+	go s.Sync()
+
+	return s.Server.Run()
+}
+
+func (s *WindowsServer) Sync() {
+	klog.Info("Sync thread started.")
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM, syscall.SIGINT)
```
bhaskardeep: >(signalChan, syscall.SIGTERM, syscall.SIGINT) [](start = 14, length = 45)

The signal channel and exit channel is not linked

pkg/mic/mic.go
bowu24: Yeah. I implement a way to avoid the change.

cmd/nmi/main.go
bowu24: With the new change, it is back in this file:)

pkg/nmi/server/server.go
bowu24: I create a interface RedirectorInt in this file and use two new files redirector_windows.go and rifirector_linux.go to impleemnt the function in this interface. In this way, we do not need the two servers. And client will call the server.run, which is the same as before.

pkg/nmi/server/linuxServer.go
bowu24: I create a interface RedirectorInt in this file and use two new files redirector_windows.go and rifirector_linux.go to impleemnt the function in this interface. In this way, we do not need the two servers and we do not have RunServer() defined twice.

pkg/pod/pod.go
bowu24: Yes, this is some potential performance issue here. Ideally ListPods should only run once when the nmi pod starts. Maybe at that time, there are not many pods existing. 
If the nmi pod restarts for some reason, it will query all the pods in the cluster, which may cause perf issue. I did some investigation and it seems we cannot query pods by specific node. I will do more investigation on it.

pkg/nmi/server/redirector_windows.go
```diff
@@ -0,0 +1,141 @@
...
+
+	for {
+		select {
+		case <-signalChan:
+			s.DeleteRoutePolicyForExistingPods()
+			break
+		case pod = <-s.Server.PodObjChannel:
+			klog.V(6).Infof("Received event: %s", pod)
+
+			fmt.Printf("Node IP and Node Name:%s %s \n", s.Server.HostIP, s.Server.NodeName)
```
alukyan: >fmt.Printf("Node IP and Node Name:%s %s \n", s.Server.HostIP, s.Server.NodeName) [](start = 3, length = 80)

This project uses `klog` not standard `fmt` for logging. You need to use existing logging approach or remove those Printf's

pkg/pod/pod.go
alukyan: So, why not do something about this. BTW: `GetPods()` method above uses `PodWatcher.Lister().List(crdSelector)`
In your code you use `PodWatcher.Informer().GetStore().List()` - why there is a difference about which API used here? I expect those queries to be different except filtering conditions.

---
In reply to: [378592628](https://github.com/Azure/aad-pod-identity/pull/496#discussion_r378592628) [](ancestors = 378592628)

pkg/nmi/server/redirector_windows.go
```diff
@@ -0,0 +1,141 @@
...
+
+type WindowsRedirector struct {
+	Server *Server
+}
+
+func makeRedirectorInt(server *Server) RedirectorInt {
+	return &WindowsRedirector{Server: server}
+}
+
+var podMap = make(map[types.UID]string)
```
alukyan: Suggest moving this map to Sync() so that access to this map is localized to Sync() method to avoid possible concurrency issues

pkg/pod/pod.go
MarcPow: I'd suggest giving this an upper limit, so we're not blocked forever in the start of an error condition.

pkg/pod/pod.go
```diff
@@ -98,6 +117,52 @@ func (c *Client) GetPods() (pods []*v1.Pod, err error) {
...
+		klog.V(6).Infof("Appending pod ID: %s/%s to list.", v1Pod.UID, v1Pod.Name)
+	}
+
+	return resList, nil
+}
+
+// GetPod returns the pod object
+func GetPod(obj interface{}, i informersv1.PodInformer) (pod *v1.Pod, err error) {
+	currentPod, exists, err := i.Informer().GetStore().Get(obj)
+	if !exists || err != nil {
```
MarcPow: Is there a potential timing condition here?  Will we ever get called while waiting for the pod to get created?

MarcPow:
@aramase Any additional thoughts here?

PR #522, opened 2020-03-04
switch to using patch instead of createorupdate

aramase:
**Reason for Change**:

- Switch to using `PATCH` instead of `CreateOrUpdate`

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/287

**Notes for Reviewers**:


pkg/cloudprovider/cloudprovider.go
```diff
@@ -179,24 +177,20 @@ func (c *Client) UpdateUserMSI(addUserAssignedMSIIDs, removeUserAssignedMSIIDs [
...
+	}
+	// remove msi ids from the list
+	for _, userAssignedMSIID := range removeUserAssignedMSIIDs {
+		ids[userAssignedMSIID] = false
 	}
+	requiresUpdate := info.SetUserIdentities(ids)
+
 	if requiresUpdate {
 		klog.Infof("Updating user assigned MSIs on %s", name)
+
```
kkmsft: nit: remove space.

cmd/demo/main.go
```diff
@@ -8,7 +8,8 @@ import (
 	"strings"
 	"time"
 
-	compute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
```
kkmsft: Could we update to the latest :  2019-12-01 or so ?

test/e2e/identityvalidator/identityvalidator.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 
-	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-04-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
```
kkmsft: could we update to the latest here as well ?

pkg/cloudprovider/vmss.go
```diff
@@ -121,41 +117,80 @@ func (h *vmssIdentityHolder) IdentityInfo() IdentityInfo {
...
 	}
-	return nil
+	return ids
 }
 
-func (i *vmssIdentityInfo) AppendUserIdentity(id string) bool {
-	if i.info.IdentityIds == nil {
-		var ids []string
-		i.info.IdentityIds = &ids
+func (i *vmssIdentityInfo) SetUserIdentities(ids map[string]bool) bool {
```
kkmsft: Can you please add a comment on what this function does ?

pkg/cloudprovider/vmss.go
```diff
@@ -121,41 +117,80 @@ func (h *vmssIdentityHolder) IdentityInfo() IdentityInfo {
...
+		i.info.UserAssignedIdentities = make(map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue)
 	}
-	return appendUserIdentity(&i.info.Type, i.info.IdentityIds, id)
-}
 
-func (i *vmssIdentityInfo) GetUserIdentityList() []string {
-	if i.info.IdentityIds == nil {
-		return []string{}
+	for id, _ := range i.info.UserAssignedIdentities {
+		expectedIDList[id] = true
```
kkmsft: nodeIDList ?

pkg/cloudprovider/vmss.go
```diff
@@ -121,41 +117,80 @@ func (h *vmssIdentityHolder) IdentityInfo() IdentityInfo {
...
+			delete(expectedIDList, id)
+		}
+		// doesn't exist on the node and want to add new identity
+		if !exists && add {
+			expectedIDList[id] = true
+		}
 	}
-	return *i.info.IdentityIds
+
+	// all identities are the node are to be removed
```
kkmsft: nit: are => on

pkg/cloudprovider/vmss.go
```diff
@@ -121,41 +117,80 @@ func (h *vmssIdentityHolder) IdentityInfo() IdentityInfo {
...
+		expectedIDList[id] = true
+	}
+
+	for id, add := range ids {
+		_, exists := expectedIDList[id]
+
+		// identity already exists/not exists, so skip doing add/del operation for
+		// the identity again for the node
+		if (exists && add) || (!exists && !add) {
+			delete(ids, id)
```
kkmsft: Is it always safe to delete the item (id) from a map(ids) which is current being iterated through ?

pkg/cloudprovider/vmss.go
```diff
@@ -121,41 +117,80 @@ func (h *vmssIdentityHolder) IdentityInfo() IdentityInfo {
...
+		}
+		// already exists on node and want to remove existing identity
```
kkmsft: Could we simplify the logic here to use either expectedIDList or ids. Is it possible to combine the logic here and the last loop and avoid to iterate again and do modifications to the ids ? And then return in case of empty node ids ?

pkg/cloudprovider/vm.go
```diff
@@ -130,33 +126,69 @@ type vmIdentityInfo struct {
...
+		if i.info.Type == compute.ResourceIdentityTypeSystemAssignedUserAssigned {
+			i.info.Type = compute.ResourceIdentityTypeSystemAssigned
+		} else {
+			i.info.Type = compute.ResourceIdentityTypeNone
+		}
+		return true
 	}
-	return *i.info.IdentityIds
+	// add and remove the new list of identities keeping the same type as before
+	userAssignedIdentities := make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
```
kkmsft: Looks like we can have common code for this and the code in vmss with just the change of the type being used to assign to the value ? Can we make that change, please ?

go.sum
```diff
@@ -191,6 +203,7 @@ golang.org/x/sys v0.0.0-20181116152217-5ac8a444bdc5/go.mod h1:STP8DvDyc/dI5b8T5h
 golang.org/x/sys v0.0.0-20181122145206-62eef0e2fa9b/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20190403152447-81d4e9dc473e/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20190412213103-97732733099d/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
```
kkmsft: Any idea how this is getting added ?

PR #539, opened 2020-03-25
set resource identity type explicitly

aramase:
**Reason for Change**:

In case of VMSS, if no identities exist the RP assumes the identity type to be `UserAssigned` if not explicitly set. This is not the case for VMAS. So this PR explicitly sets the type based on the current resource identity type.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/identity.go
+++ b/pkg/cloudprovider/identity.go
@@ -32,8 +32,9 @@ func checkIfIDInList(idList []string, desiredID string) bool {
 	return false
 }
 
-// getResourceIdentityType returns resource identity type based on current type
-func getResourceIdentityType(identityType compute.ResourceIdentityType) compute.ResourceIdentityType {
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/identity.go
```diff
@@ -29,3 +31,15 @@ func checkIfIDInList(idList []string, desiredID string) bool {
 	}
 	return false
 }
+
+// getResourceIdentityType returns resource identity type based on current type
+func getResourceIdentityType(identityType compute.ResourceIdentityType) compute.ResourceIdentityType {
```
kkmsft: Could you please elaborate in the comment how the flow of code brings us to the state where we have to check the updated resource type based on the current resource type. Does function name like - getUpdatedIdentityType or something like that which indicates the nature of function make more sense ?

Issue #533, opened 2020-03-18
MIC doesn't delete user managed identities on namespace deletion


arodriguezdlc:
**Describe the bug**
As it's explained at documentation (https://github.com/Azure/aad-pod-identity#node-managed-identity), When AzureIdentity resource or all pods with this AzureIdentity assigned are deleted, MIC should automatically remove the User Managed Identity from VMSS. 

This works well if you delete directly this resources, but fails if the kubernetes namespace with this resources is deleted. MIC should notice that this resource doesn't exist anymore and should delete the user managed identity association with vmss. 

In our case, this issue is triggering this other issue: https://github.com/Azure/aad-pod-identity/issues/515. When we destroy the k8s namespace, we also delete User managed identity, but due to MIC isn't cleaning the association with the vmss, when MIC tries to create a new association, it fails in the CreateOrUpdate operation because there is an association with an unexisting user managed identity. 

**Steps To Reproduce**
1. Create a User Managed Identity
1. Create a namespace
```
kubectl create namespace test
````
1. Create an Azure Identity and an Azure Identity Binding. Azure Identity must be configured with
resource id and client id from User Managed Identity. 
```yaml
apiVersion: "aadpodidentity.k8s.io/v1"
kind: AzureIdentity
metadata:
  name: test
  namespace: test
  annotations:
    aadpodidentity.k8s.io/Behavior: namespaced
spec:
  type: 0
  ResourceID: # From User Managed Identity
  ClientID: # From User Managed Identity
---
apiVersion: "aadpodidentity.k8s.io/v1"
kind: AzureIdentityBinding
metadata:
  name: test
  namespace: test
spec:
  AzureIdentity: test
  Selector: test
``` 
1. Create a pod with the selector
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  namespace: test
  labels:
    app: myapp
    aadpodidbinding: test
spec:
  containers:
  - name: myapp-container
    image: busybox
    command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 3600']
```

Now you can check that user managed identity should be associated to the vmss.

1. Delete namespace
```
kubectl delete namespace test
```

User managed identity is not deleted. 
**Expected behavior**
User managed identity association with vmss should be deleted if associated k8s resources are deleted. 

**AAD Pod Identity version**
1.5.5

**Kubernetes version**
1.15.7 (AKS)

**Additional context**
[...]

Issue #504, opened 2020-02-11
Server reported a conflict


kwaazaar:
**Describe the bug**
The longer I try to use AAD Pod Identity, the less it seems to work. I'm now seeing these MIC-related errors.

kubectl logs -l component=mic -n default
~~~
E0211 14:02:59.717474       1 crd.go:360] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:02:59.717597       1 mic.go:814] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
I0211 14:02:59.763439       1 cloudprovider.go:199] Updating user assigned MSIs on aks-agentpool-71662594-17
E0211 14:03:00.111814       1 crd.go:360] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:03:00.111867       1 mic.go:814] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:03:00.433394       1 crd.go:360] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:03:00.433446       1 mic.go:814] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:03:00.517374       1 crd.go:360] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
E0211 14:03:00.517428       1 mic.go:814] the server reported a conflict (post azureassignedidentities.aadpodidentity.k8s.io)
I0211 14:03:00.554095       1 cloudprovider.go:199] Updating user assigned MSIs on aks-agentpool-71662594-9
E0205 13:23:57.339073       1 leaderelection.go:224] error retrieving resource lock default/aad-pod-identity-mic: Get https://aahnl-kubernetes-ot-78c6386c.hcp.westeurope.azmk8s.io:443/api/v1/namespaces/default/endpoints/aad-pod-identity-mic: EOF
E0205 13:24:13.536585       1 leaderelection.go:224] error retrieving resource lock default/aad-pod-identity-mic: Get https://aahnl-kubernetes-ot-78c6386c.hcp.westeurope.azmk8s.io:443/api/v1/namespaces/default/endpoints/aad-pod-identity-mic: net/http: TLS handshake timeout
~~~

Not sure if related, but this is what the nmi logs state:
~~~
[...]

PR #546, opened 2020-03-30
add finalizer for azure assigned identity

aramase:
**Reason for Change**:

- Add finalizer to prevent manual deletion of `AzureAssignedIdentity` before the identities are removed from underlying vm/vmss.
- `PUT` first to update existing `AzureAssignedIdentity`. If not exists, then `POST`.

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/533, fixes https://github.com/Azure/aad-pod-identity/issues/504

**Notes for Reviewers**:

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -10,6 +10,7 @@ import (
 	aadpodv1 "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
 	"github.com/Azure/aad-pod-identity/pkg/metrics"
 	"github.com/Azure/aad-pod-identity/pkg/stats"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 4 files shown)


pkg/crd/crd.go
```diff
@@ -427,10 +432,19 @@ func (c *Client) RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureAssigned
...
-
 	}()
 
-	err = c.rest.Delete().Namespace(assignedIdentity.Namespace).Resource("azureassignedidentities").Name(assignedIdentity.Name).Do().Error()
+	var res aadpodv1.AzureAssignedIdentity
+	err = c.rest.Delete().Namespace(assignedIdentity.Namespace).Resource(aadpodid.AzureAssignedIDResource).Name(assignedIdentity.Name).Do().Into(&res)
+	if err != nil && apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err == nil && HasFinalizer(res.ObjectMeta) {
```
kkmsft, line 442: why do we have to check if there is a finalizer. Are there cases where it could not be there. Is that for backward compat ? If so could you please add a comment.

aramase: There are 2 reasons it's possible -
1. It's an old azure assigned identity
2. The finalizer was manually removed

This check is just to see if we need to remove the finalizer if exists, so we can unblock garbage collection of deleted azureassignedidentity. Will add this as a comment.

aramase:
Closing this PR. Will reopen only with finalizer changes.

Issue #466, opened 2020-01-02
Failed to Assign User Managed Identity to VM


droessmj:
**Describe the bug**
Multiple User Managed Identities were not bound to their node VMs  due to incorrect configuration for one specific AzureIdentity resource on a cluster containing multiple AzureIdentities across multiple namespaces. 

**Steps To Reproduce**
Create and bind an incorrectly configured AzureIdentity resource with a null ClientId and ResourceId. This resource can be created in any namespace. 

Create and bind a correctly configured AzureIdentity while the incorrectly configured resource exists on the cluster. 

The correctly configured AzureIdentity will have the AzureAssignedIdentity resource created, but the User Managed Identity will not be bound to the VM under the hood. NMI logs will show that the AssignedIdentity cannot be found on the VM. 

**Expected behavior**
Configuration of one AzureIdentity does not impact the ability of other AzureIdentity resources to associate or disassociate with a node VM. 

**AAD Pod Identity version**
1.5.2

**Kubernetes version**
1.14.8

**Additional context**
The behavior appears to demonstrate that identity assignment is being performed en masse or as a batch, with one failure failing the entire operation. MIC logs for the "correctly configured" resource indicated incorrect values, which I eventually tracked down to the other, "incorrectly configured" resource. After removing the incorrect AzureIdentity from its namespace, all User Managed Identities were correctly assigned to their respective VM nodes. 

**NMI Log Example**
time="2020-01-02T15:22:49Z" level=error msg="no AzureAssignedIdentity found for pod:PODNAME_REPLACED in assigned state, context canceled" req.method=GET req.path=/metadata/identity/oauth2/token req.remote=10.244.2.21

**MIC Logs**
E0102 15:00:17.050558 1 vm.go:48] compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"LinkedInvalidPropertyId\" Message=\"Property id '' at path 'identity.identityIds[2]' is invalid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'.\" ","
[...]

PR #548, opened 2020-04-01
add validation for resource id

aramase:
**Reason for Change**:

Validate `ResourceID` from `AzureIdentity` and skip processing `AzureIdentity` if resource id is not valid. This prevents `LinkedInvalidPropertyId` errors during vm/vmss `UPDATE` calls with invalid identities. 

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/466

**Notes for Reviewers**:

### DIFF
--- a/pkg/mic/mic.go
+++ b/pkg/mic/mic.go
@@ -15,6 +15,7 @@ import (
 	"github.com/Azure/aad-pod-identity/pkg/metrics"
 	"github.com/Azure/aad-pod-identity/pkg/pod"
 	"github.com/Azure/aad-pod-identity/pkg/stats"
+	"github.com/Azure/aad-pod-identity/pkg/utils"
 	"github.com/Azure/aad-pod-identity/version"
 	"golang.org/x/sync/semaphore"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/mic/mic.go
```diff
@@ -815,10 +816,16 @@ func getIDKey(ns, name string) string {
...
-	m = make(map[string]aadpodid.AzureIdentity, len(arr))
-	for _, element := range arr {
-		m[getIDKey(element.Namespace, element.Name)] = element
+func (c *Client) convertIDListToMap(azureIdentities []aadpodid.AzureIdentity) (m map[string]aadpodid.AzureIdentity, err error) {
+	m = make(map[string]aadpodid.AzureIdentity, len(azureIdentities))
+	for _, azureIdentity := range azureIdentities {
+		// validate the resourceID in azure identity to ensure format is as expected
+		err := utils.ValidateResourceID(azureIdentity.Spec.ResourceID)
+		if err != nil {
+			klog.Errorf("Ignoring azure identity %s/%s, error: %v", azureIdentity.Namespace, azureIdentity.Name, err)
```
kkmsft, line 825: Should this be a warning ?

pkg/mic/mic_test.go
```diff
@@ -20,13 +20,17 @@ import (
...
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 )
 
+var (
+	testResourceID = "/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity1"
```
kkmsft, line 31: Could we add set of negative test cases with wrong format of resource id ?

pkg/utils/utils_test.go
```diff
@@ -18,9 +18,45 @@ func TestRedactClientID(t *testing.T) {
...
+		expectedErr bool
+	}{
+		{
+			name:        "invalid resource id 0",
+			resourceID:  "invalidresid",
+			expectedErr: true,
+		},
+		{
+			name:        "invalid resource id 1",
+			resourceID:  "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourcegroups/0000/providers/Microsoft.ManagedIdentity/keyvault-identity-0",
```
kkmsft, line 43: Good that we have test at this level, could we add it one level above as well since there is some logic there ?

pkg/mic/mic.go
aramase: I think we should keep it as an error. If users are filtering logs based on errors, this should surface out as to why the pod didn't get an `AzureAssignedIdentity`.

pkg/mic/mic_test.go
aramase: The negative test case is already in here - https://github.com/Azure/aad-pod-identity/pull/548/files#diff-33233ef23f7b38ee10add9aed587cd05R770-R775.

pkg/utils/utils_test.go
aramase: There is test one level above at MIC

PR #558, opened 2020-04-10
type upgrade azure assigned identity

aramase:
**Reason for Change**:

- Upgrade `AzureAssignedIdentity` individually using upgraded `AzureIdentity` and `AzureIdentityBinding` to prevent any reloads.

**Issue Fixed**:

**Notes for Reviewers**:


pkg/crd/crd.go
```diff
@@ -362,6 +361,74 @@ func (c *Client) UpgradeAll() error {
...
+	list, err := meta.ExtractList(i)
+	if err != nil {
+		return fmt.Errorf("extract list error for resource: %s, err: %v", aadpodv1.AzureIDResource, err)
+	}
+	for _, item := range list {
+		o, err := meta.Accessor(item)
+		if err != nil {
+			return fmt.Errorf("get object for resource: %s, error: %v", aadpodv1.AzureIDResource, err)
+		}
+		updatedAzureIdentities[o.GetNamespace()+"/"+o.GetName()] = *o.(*aadpodv1.AzureIdentity)
```
kkmsft, line 380: Can we populate the cache directly when doing the update of the AzureIdentity and Binding instead of doing another fetch and parse again. Currently the "Into" of the setObject is not used. We can use that fill the cache avoiding the second set of calls.

pkg/crd/crd.go
```diff
@@ -362,6 +361,74 @@ func (c *Client) UpgradeAll() error {
...
+		if err != nil {
+			return fmt.Errorf("get object for resource: %s, error: %v", aadpodv1.AzureAssignedIDResource, err)
+		}
+		obj := o.(*aadpodv1.AzureAssignedIdentity)
+		idName := obj.Spec.AzureIdentityRef.Name
+		idNamespace := obj.Spec.AzureIdentityRef.Namespace
+		bindingName := obj.Spec.AzureBindingRef.Name
+		bindingNamespace := obj.Spec.AzureBindingRef.Namespace
+
+		if v, exists := updatedAzureIdentities[idNamespace+"/"+idName]; exists {
```
kkmsft, line 421: Can you please add a short util function to this append and use namespace and name as inputs ?

Issue #484, opened 2020-01-16
MIC can't get access to Azure Instance Metadata service


Kuzbekov:
When we use MSI for kubernetes, MIC can't update VM's because its requests catched by NMI.
Even if I label that pod with correct label it still can't assign Identity to itself.
Temporary killing NMI DaemonSet unlock MIC and allows it to correctly do all assignments.

> [aad-pod-identity-mic-f4c46b576-vc5h6] I0116 07:50:47.348141       1 mic.go:787] Processing node usazuenade00346, add [44], del [44] 
> [aad-pod-identity-mic-f4c46b576-vc5h6] E0116 07:52:07.433303       1 vmss.go:103] azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/REDACTED/resourceGroups/REDACTED/providers/Microsoft.Compute/virtualMachineScaleSets/REDACTED_scalesetname?api-version=2018-04-01: StatusCode=403 -- Original Error: adal: Refresh request failed. Status Code = '403'. Response body: no AzureAssignedIdentity found for pod:default/aad-pod-identity-mic-f4c46b576-vc5h6 in assigned state 
> [aad-pod-identity-mic-f4c46b576-vc5h6] E0116 07:52:07.433367       1 mic.go:831] Updating msis on node REDACTED_scalesetname, add [44], del [44] failed with error azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/REDACTED/resourceGroups/REDACTED/providers/Microsoft.Compute/virtualMachineScaleSets/REDACTED_scalesetname?api-version=2018-04-01: StatusCode=403 -- Original Error: adal: Refresh request failed. Status Code = '403'. Response body: no AzureAssignedIdentity found for pod:default/aad-pod-identity-mic-f4c46b576-vc5h6 in assigned state 

**Steps To Reproduce**
Deploy Kubernetes with AKS-engine using  Managed Identity account.
Deploy aad-pod-identity with helm chart from this repository

**Expected behavior**
requests to 169.254.169.254 from mic pod should be passed to Azure Instance Metadata service rather then catched by nmi

**AAD Pod Identity version**
1.5.4

**Kubernetes version**
1.14.7 Deployed with AKS-Engine Using Managed Identity

kkmsft:
Thank you for using aad pod identity @Kuzbekov . 
[...]

PR #561, opened 2020-04-10
add mic pod exception to helm chart

aramase:
**Reason for Change**:

When NMI runs on the same node as MIC pod, then MIC request to fetch token is intercepted by NMI. This token request will fail pod-identity validation as MIC is yet to initialize and NMI can't find an `AzureAssignedIdentity` for MIC. With this change, we install the `AzurePodIdentityException` for MIC pod labels so token requests from MIC go directly to IMDS instead of pod-identity validation path.

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/484

**Notes for Reviewers**:


charts/aad-pod-identity/values.yaml
```diff
@@ -14,6 +14,13 @@ image:
 # To match only pods in the namespace containing AzureIdentity set this to true.
 forceNameSpaced: "false"
 
+# When NMI runs on a node where MIC is running, then MIC token request call is also
+# intercepted by NMI. MIC can't get a valid token as to initialize and then
+# assign the identity. Installing an exception for MIC would ensure all token requests
+# for MIC pods directly go to IMDS and not go through the pod-identity validation
+# https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.app-exception.md
+installMICException: "true"
```
kkmsft, line 22: Is this required by default ? Wouldn't this only be for system assigned identity clusters ?

charts/aad-pod-identity/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Deploy components for aad-pod-identity
 name: aad-pod-identity
-version: 1.5.5
+version: 1.5.6
```
kkmsft, line 4: Is this going to be updated right now or near the release ?

charts/aad-pod-identity/values.yaml
aramase: This is required for all helm installs for AKS clusters. In AKS clusters, NMI runs on the same node as MIC and fails to fetch tokens.

charts/aad-pod-identity/Chart.yaml
aramase: This is just chart version. I can make a tar file in this PR or create a new one with that? wdyt?

PR #564, opened 2020-04-15
process del id list before add

aramase:
**Reason for Change**:

Add is done after setting del ids in the map to ensure an identity if in both add and del list is not deleted

**Issue Fixed**:

**Notes for Reviewers**:


pkg/cloudprovider/cloudprovider.go
```diff
@@ -178,14 +178,16 @@ func (c *Client) UpdateUserMSI(addUserAssignedMSIIDs, removeUserAssignedMSIIDs [
...
-		ids[userAssignedMSIID] = true
-	}
 	// remove msi ids from the list
 	for _, userAssignedMSIID := range removeUserAssignedMSIIDs {
 		ids[userAssignedMSIID] = false
 	}
+	// add new ids to the list
+	// add is done after setting del ids in the map to ensure an identity if in
+	// both add and del list is not deleted
+	for _, userAssignedMSIID := range addUserAssignedMSIIDs {
```
kkmsft, line 188: Do we have UTs specific to this scenario ?

aramase: Yeah, it's added as part of this PR!

PR #572, opened 2020-04-22
chore: add docs for v1.6.0 breaking change and troubleshooting guide

chewong:
**Reason for Change**:

Add documentation regarding the breaking change introduced in v1.6.0 and a troubleshooting guide for aad-pod-identity.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Using Kubernetes primitives, administrators configure identities and bindings to
 
 ## v1.6.0 Breaking Change
 
-With https://github.com/Azure/aad-pod-identity/pull/398, the [client-go](https://github.com/kubernetes/client-go) library is upgraded to v0.17.2, where CRD fields are now case sensitive. If you are upgrading MIC and NMI from v1.x.x to v1.6.0, no action is required since existing `AzureIdentity` and `AzureIdentityBinding` are backward-compatible.
+With https://github.com/Azure/aad-pod-identity/pull/398, the [client-go](https://github.com/kubernetes/client-go) library is upgraded to v0.17.2, where CRD fields are now case sensitive. If you are upgrading MIC and NMI from v1.x.x to v1.6.0, MIC v1.6.0+ will upgrade the fields of existing `AzureIdentity` and `AzureIdentityBinding` on startup to the new format to ensure backward compatibility. A configmap called `aad-pod-identity-config` is created to record and confirm the successful type upgrade.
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -27,6 +28,28 @@ Using Kubernetes primitives, administrators configure identities and bindings to
 * [What To Do Next?](#what-to-do-next)
 * [Code of Conduct](#code-of-conduct)
 
+## v1.6.0 Breaking Change
+
+With https://github.com/Azure/aad-pod-identity/pull/398, the [client-go](https://github.com/kubernetes/client-go) library is upgraded to v0.17.2, where CRD fields are now case sensitive. If you are upgrading MIC and NMI from v1.x.x to v1.6.0, no action is required since existing `AzureIdentity` and `AzureIdentityBinding` are backward-compatible.
```
aramase, line 33: Also, one more note to add - For existing `AzureIdentity` and `AzureIdentityBinding`, MIC will convert all the **existing resources** on startup to the new format and on successfully doing that will write to `aad-pod-identity-config` configmap

README.md
```diff
@@ -27,6 +28,28 @@ Using Kubernetes primitives, administrators configure identities and bindings to
...
+
+However, for future `AzureIdentity` and `AzureIdentityBinding` created using v1.6.0+, the following fields need to be changed:
```
aramase, line 35: We should also include [AzurePodIdentityException](https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.app-exception.md) CRD and describe the `PodLabels` to `podLabels` change for the field

docs/readmes/README.troubleshooting.md
```diff
@@ -0,0 +1,41 @@
...
+```bash
+kubectl logs -l component=mic --since=1h | grep ^E
+kubectl logs -l component=nmi --since=1h | grep ^E
+```
+
+> It is always a good idea to include relevant logs from MIC and NMI when opening a new [issue](https://github.com/Azure/aad-pod-identity/issues).
+
+### Ensure that iptables rule exists
+
+To ensure that the correct iptables rule is injeceted to each node via the [NMI](../../README.md#node-managed-identity) pods, the following command ensures that on a given node, there exists an iptables rule where all packets with a destination IP of 169.254.169.254 (IMDS endpoint) are routed to port 2579 of the host network.
```
aramase, line 20: s/injeceted/injected

Issue #588, opened 2020-04-29
Helm install complains: skipping unknown hook: "crd-install"


lukerogers:
**Describe the bug**
When deploying the Helm chart using 3.1.2, it complains with the message:
> manifest_sorter.go:192: info: skipping unknown hook: "crd-install"

The chart does install, but the errors cause Azure deployments to fail.

**Steps To Reproduce**
`helm upgrade --namespace kube-system --install --set forceNameSpaced=true --wait aad-pod-identity aad-pod-identity/aad-pod-identity`

**Expected behavior**
Chart installed with no warnings.

**AAD Pod Identity version**
1.6.0

**Kubernetes version**
1.15.10

**Additional context**
Helm 3 changes the way CRD's are handled and the hook is no longer required. Resources in the /crds folder are automatically applied. I will submit a pull request for this.

lukerogers:
This breaks backwards compatibility, closing to avoid confusion.

PR #589, opened 2020-04-29
Removing legacy CRD deployment for Helm charts.

lukerogers:
**Reason for Change**:

Fixes Helm deploy to handle CRD's correctly

**Issue Fixed**:

#588

**Notes for Reviewers**:
Any feedback would be great.


lukerogers:
Excellent point, in my selfish quest to get rid of warnings, I forgot about backwards compatibility. I'll close this and just tell my deployment to ignore them. Thanks for the feedback!

PR #593, opened 2020-05-01
feat: add finalizer for assigned identity

aramase:
**Reason for Change**:

- Adds finalizer to `AzureAssignedIdentity`
  - ensures, the `AzureAssignedIdentity` is only deleted after the identity is successfully un-assigned from the node.
  - ensures identities are cleaned up when namespace is deleted.

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/533

**Notes for Reviewers**:

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -11,6 +11,7 @@ import (
 	aadpodv1 "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
 	"github.com/Azure/aad-pod-identity/pkg/metrics"
 	"github.com/Azure/aad-pod-identity/pkg/stats"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -22,6 +23,10 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+)
+
+const (
+	finalizerName = "azureassignedidentity.finalizers.aadpodidentity.k8s.io"
 )
 
 // Client represents all the watchers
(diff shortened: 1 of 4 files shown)


pkg/mic/mic.go
```diff
@@ -1096,24 +1096,19 @@ func (c *Client) updateUserMSI(newAssignedIDs map[string]aadpodid.AzureAssignedI
...
 				continue
 			}
 
-			klog.Infof("Updating msis on node %s failed, but identity %s has successfully been removed from node", delID.Spec.NodeName, removedBinding.Name)
+			klog.Infof("Updating msis on node %s failed, but identity %s/%s has successfully been removed from node", delID.Spec.NodeName, removedBinding.Namespace, removedBinding.Name)
 
 			// remove assigned identity crd from cluster as the identity has successfully been removed from the node
 			err = c.removeAssignedIdentity(&delID)
 			if err != nil {
-				c.EventRecorder.Event(removedBinding, corev1.EventTypeWarning, "binding remove error",
```
aramase, line 1109: Removing these events on binding because it's possible we are deleting the assigned identity as the binding has been deleted.

pkg/mic/mic.go
```diff
@@ -1096,24 +1096,19 @@ func (c *Client) updateUserMSI(newAssignedIDs map[string]aadpodid.AzureAssignedI
 			// the identity still exists on node, which means removing the identity from the node failed
 			if isUserAssignedMSI && !inUse && idExistsOnNode {
 				message := fmt.Sprintf("Binding %s removal from node %s for pod %s resulted in error %v", removedBinding.Name, delID.Spec.NodeName, delID.Spec.Pod, err.Error())
-				c.EventRecorder.Event(removedBinding, corev1.EventTypeWarning, "binding remove error", message)
 				klog.Error(message)
 				continue
 			}
 
-			klog.Infof("Updating msis on node %s failed, but identity %s has successfully been removed from node", delID.Spec.NodeName, removedBinding.Name)
+			klog.Infof("Updating msis on node %s failed, but identity %s/%s has successfully been removed from node", delID.Spec.NodeName, removedBinding.Namespace, removedBinding.Name)
```
chewong, line 1103: Why are we logging the name of the identity with `removdBinding`? Shouldn't we use `id`?

pkg/crd/crd.go
```diff
@@ -483,7 +488,21 @@ func (c *Client) RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureAssigned
 
 	}()
 
-	err = c.rest.Delete().Namespace(assignedIdentity.Namespace).Resource("azureassignedidentities").Name(assignedIdentity.Name).Do().Error()
+	var res aadpodv1.AzureAssignedIdentity
+	err = c.rest.Delete().Namespace(assignedIdentity.Namespace).Resource("azureassignedidentities").Name(assignedIdentity.Name).Do().Into(&res)
+	// assigned identity doesn't exist
+	if err != nil && apierrors.IsNotFound(err) {
```
chewong, line 494: nit: `if apierrors.IsNotFound(err)` should be sufficient.

pkg/crd/crd.go
```diff
@@ -483,7 +488,21 @@ func (c *Client) RemoveAssignedIdentity(assignedIdentity *aadpodid.AzureAssigned
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if hasFinalizer(&res) {
+		removeFinalizer(&res)
+		// update the assigned identity without finalizer and resource will be garbage collected
+		err = c.rest.Put().Namespace(assignedIdentity.Namespace).Resource(aadpodid.AzureAssignedIDResource).Name(assignedIdentity.Name).Body(&res).Do().Error()
```
chewong, line 503: Let's check if the error is nil or not.

chewong: Just realized you are returning the error

pkg/mic/mic.go
```diff
@@ -1080,6 +1080,7 @@ func (c *Client) updateUserMSI(newAssignedIDs map[string]aadpodid.AzureAssignedI
 
 		for _, delID := range nodeTrackList.assignedIDsToDelete {
 			id := delID.Spec.AzureIdentityRef
+			removedID := delID.Spec.AzureIdentityRef
```
chewong, line 1083: `id` and `removedID` are duplicated

PR #579, opened 2020-04-24
test: set up e2e test pipeline on ADO

chewong:
**Reason for Change**:

Adding e2e tests on ADO as a PR gate to reduce regressions introduced by PRs.

**Issue Fixed**:

Fixes #568 

**Notes for Reviewers**:

PR #596, opened 2020-05-01
fix: fix e2e test regression caused by #579

chewong:
**Reason for Change**:

#579 introduced `IDENTITY_RESOURCE_GROUP` and `CLUSTER_RESOURCE_GROUP` as part of the envconfig for e2e test, so we can no longer assume that the user-assigned identity is within the same resource group as the cluster. Instead of using the name of the user-assigned identity to identify them, we started using the resource ID obtained from `az identity show -g <IdentityResourceGroup> -n <IdentityName>`.

The resource ID obtained from `az identity show` has a `resourcegroup` part of it instead of `resourceGroup` for some unknown reasons. It caused `should not alter the user assigned identity on VM after AAD pod identity is created and deleted` test case to fail since it constructs on a map of user-assigned identity where the key is the resource ID of the identity and the test case assumes that `resourceGroup` is part of the resource ID. This PR converts all keys in the map to lower cases to mitigate the issue.

**Issue Fixed**:

https://upstream.jenkins.azure-containers.io/view/aad-pod-identity/job/aad-pod-identity-e2e-daily/

**Notes for Reviewers**:

### DIFF
--- a/test/common/azure/azure.go
+++ b/test/common/azure/azure.go
@@ -315,7 +315,7 @@ func GetVMUserAssignedIdentities(resourceGroup, vmName string) (map[string]UserA
 		return nil, errors.Wrap(err, "Failed to unmarshall json")
 	}
 
-	return userAssignedIdentities, nil
+	return convertKeysToLower(userAssignedIdentities), nil
 }
 
 // GetVMSSUserAssignedIdentities will return the list of user assigned identity in a given VM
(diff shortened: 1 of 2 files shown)


aramase:
Certain times, there are differences in case in `az vm identity show` output. In the code path, we convert to lower case to handle this scenario. I believe it's some changes done in CRP when assigning the identity object.

Issue #571, opened 2020-04-22
Enable linting in CI


aramase:
- [ ] Code clean up
- [ ] Enable linting in CI

aramase:
@elsesiy Thank you for adding the lint tools. Are you interested in making the changes to fix the lint errors?

elsesiy:
@aramase Yes I started working on that a while ago already

PR #597, opened 2020-05-02
Code clean up

elsesiy:
Fix typos
Fix swallowed errors
Report Metrics reporter errors
Remove unused/dead code
Properly goimports/gofmt files
Rename stats.StatsType to stats.Type

Issue: #571


elsesiy:
@aramase @chewong As per our conversation, I fixed the linting errors. Please have a look and let me know if you want me to undo some of the fixes, thanks!

pkg/cloudprovider/cloudprovider_test.go
```diff
@@ -5,11 +5,12 @@ import (
...
 	"testing"
 
-	"github.com/Azure/aad-pod-identity/pkg/config"
 	"github.com/Azure/go-autorest/autorest/azure"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/Azure/aad-pod-identity/pkg/config"
+
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
```
chewong, line 14: This import should be grouped with the external import group.

pkg/auth/auth.go
```diff
@@ -22,10 +24,16 @@ func GetServicePrincipalTokenFromMSI(resource string) (*adal.Token, error) {
 
 	defer func() {
 		if err != nil {
-			reporter.ReportIMDSOperationError(metrics.AdalTokenFromMSIOperationName)
+			err = reporter.ReportIMDSOperationError(metrics.AdalTokenFromMSIOperationName)
+			if err != nil {
+				klog.Warningf("Metrics reporter error: %+v", err)
```
chewong, line 29: Since we are logging an error, we should use `klog.Errorf` instead of `klog.Warningf`.

pkg/cloudprovider/cloudprovider_test.go
```diff
@@ -76,20 +77,41 @@ func TestSimple(t *testing.T) {
...
 			node4 := &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node4-vmss0000000"}, Spec: corev1.NodeSpec{ProviderID: vmssProvider}}
 
-			cloudClient.UpdateUserMSI([]string{"ID0", "ID0again"}, []string{}, node0.Name, false)
-			cloudClient.UpdateUserMSI([]string{"ID1"}, []string{}, node1.Name, false)
-			cloudClient.UpdateUserMSI([]string{"ID2"}, []string{}, node2.Name, false)
-			cloudClient.UpdateUserMSI([]string{"ID3"}, []string{}, node3.Name, false)
-			cloudClient.UpdateUserMSI([]string{"ID4"}, []string{}, node4.Name, true)
+			err := cloudClient.UpdateUserMSI([]string{"ID0", "ID0again"}, []string{}, node0.Name, false)
+			if err != nil {
+				t.Error("Couldn't update MSI")
```
chewong, line 82: We should use `t.Errorf` to log the error message too.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -3,22 +3,22 @@ package cloudprovider
...
-	"github.com/Azure/aad-pod-identity/pkg/utils"
-	"github.com/Azure/aad-pod-identity/version"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
 	yaml "gopkg.in/yaml.v2"
 	"k8s.io/klog"
+
+	config "github.com/Azure/aad-pod-identity/pkg/config"
```
chewong, line 19: Pacakge renaming shouldn't be necessary here.

pkg/nmi/standard.go
```diff
@@ -6,11 +6,12 @@ import (
...
 
+	"github.com/Azure/go-autorest/autorest/adal"
+
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
 	auth "github.com/Azure/aad-pod-identity/pkg/auth"
 	k8s "github.com/Azure/aad-pod-identity/pkg/k8s"
 	utils "github.com/Azure/aad-pod-identity/pkg/utils"
-	"github.com/Azure/go-autorest/autorest/adal"
 
 	"k8s.io/klog"
```
chewong, line 16: This import should be grouped with the external import group as well.

cmd/mic/main.go
```diff
@@ -8,13 +8,14 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/client-go/rest"
```
chewong, line 11: This repo has been following the import order of standard libraries -> local packages (`github.com/Azure/aad-pod-identity/...`) -> external packages. Let's keep this convention and add whitespace between those three import groups.

pkg/cloudprovider/cloudprovider_test.go
elsesiy: I don't know what you're asking, I ran `goimports` on the import path which is something that the linter complains about

pkg/auth/auth.go
elsesiy: These errors are related to metrics submission and not really the as severe as an application error so I thought I just log a warning instead but I can change it if you want

pkg/cloudprovider/cloudprovider_test.go
chewong: `github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute` should be above the `github.com/Azure/go-autorest/autorest/azure` import but `goimports` didn't catch it probably because it has a newline above it.

pkg/auth/auth.go
chewong: Makes sense, let's keep it as it is.

pkg/auth/auth.go
```diff
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"time"
 
+	adal "github.com/Azure/go-autorest/autorest/adal"
```
aramase, line 7: we can probably remove the package name for this import - `"github.com/Azure/go-autorest/autorest/adal"` should be sufficient

pkg/cloudprovider/cloudprovider_test.go
elsesiy: Okay I'm going to get rid of the newlines and re-run `goimports`.

elsesiy:
@aramase @chewong Updated the PR with the review comments, please let me know if it's good to go now

.pipelines/unit-tests-template.yml
```diff
@@ -12,5 +12,7 @@ jobs:
         displayName: Build
       - script: make unit-test
         displayName: Run unit tests
+      - script: make lint
```
chewong, line 15: Could you increase the timeout from 10 min to maybe 20 minutes? Looks like it timed out after 10 minutes.

pkg/crd/crd_test.go
```diff
@@ -4,11 +4,11 @@ import (
 	"testing"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/rest"
+
```
chewong, line 7: nit: Remove extra space.

cmd/mic/main.go
chewong: @elsesiy  Thanks for addressing my comment. Could you also apply the same change to other files?

.pipelines/unit-tests-template.yml
elsesiy: Sure, just wanted to point out that the 10m cap applies to the job and not the step. The current run failed after about 6min.

cmd/mic/main.go
elsesiy: Done! Just wondering why that matters since you could just always `goimports` the files and have it done the "right" way?

chewong: Thanks for working on this. Linting tools, in general, are subjective and I don't think there is an absolute "right" way to lint a file. Although I agree with almost all of the changes in this PR, I think we should also follow some conventions laid down by the repository.

elsesiy:
> Hmm golangci-lint v1.24.0 hangs in Azure DevOps pipeline. What do you think of upgrading it to v1.27.0? I opened a test PR #616 and it looks like it works with v1.27.0.

I was skeptical already with the long build time so that might explain it, I'll update it.

cmd/mic/main.go
elsesiy: Thanks for the explanation & I agree there's no single way to lint a file. I was just curious to understand why the repo chose this import ordering.

elsesiy:
@chewong I fixed the last two files 👍

elsesiy:
Thank you!

PR #601, opened 2020-05-05
feat: make update user msi calls retriable

### LINKED ISSUE #599: Add retries with updating identities
**Describe the request**
If there is an error while assigning/un-assigning identity from the node, ARM returns the type of error and the identity that has the issue. We should check if the error is of known type (eg. `LinkedAuthorizationFailed`), get the identity in error message, remove it from the list and retry identity assignment again with the new list.

Identities are assigned/un-assigned from the underlying node in batches. This means if there is an error with a single identity, the entire UPDATE operation is rejected. So a single misconfigured identity could result in failure to assign any further identities until the identity in issue is fixed. 

**Additional context**

chewong:
**Reason for Change**:

Attempts to extract erroneous user-assigned identities from ARM error message and remove them from the identity update list before retrying.

**Issue Fixed**:

Fixes #599.

**Notes for Reviewers**:
Steps I took to test:
1. Create two AzureIdentities and AzureIdentityBindings, one with valid resource ID and one with resource ID that does not exist
2. Create a pod that consumes the two AzureIdentities
3. Check that there are two AzureAssignedIdentities created. The one with valid resource ID should be in "Assigned" state and the other should be in "Created" state.
3. Check that one of the user-assigned identities is assigned correctly to the underlying node on Azure, and the other one isn't.

### DIFF
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -34,6 +34,7 @@ var (
 	immutableUserMSIs   string
 	cmConfig            mic.CMConfig
 	typeUpgradeConfig   mic.TypeUpgradeConfig
+	updateUserMSIConfig mic.UpdateUserMSIConfig
 )
 
 func main() {
@@ -81,6 +82,10 @@ func main() {
 	// Config map details for the type changes in the context of client-go upgrade.
 	flag.StringVar(&typeUpgradeConfig.TypeUpgradeStatusKey, "type-upgrade-status-key", "type-upgrade-status", "Configmap key for type upgrade status")
 	flag.BoolVar(&typeUpgradeConfig.EnableTypeUpgrade, "enable-type-upgrade", true, "Enable type upgrade")
+
+	// Parameters for retrying cloudprovider's UpdateUserMSI function
+	flag.IntVar(&updateUserMSIConfig.MaxRetry, "update-user-msi-max-retry", 2, "The maximum retry of UpdateUserMSI call")
+	flag.DurationVar(&updateUserMSIConfig.RetryPeriod, "update-user-msi-retry-period", 1*time.Second, "The duration to wait before retrying UpdateUserMSI")
 
 	flag.Parse()
 
@@ -142,6 +147,7 @@ func main() {
 		ImmutableUserMSIsList: immutableUserMSIsList,
 		CMcfg:                 &cmConfig,
 		TypeUpgradeCfg:        &typeUpgradeConfig,
+		UpdateUserMSICfg:      &updateUserMSIConfig,
 	}
 
 	micClient, err := mic.NewMICClient(micConfig)
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -85,7 +85,7 @@ func main() {
 
 	// Parameters for retrying cloudprovider's UpdateUserMSI function
 	flag.IntVar(&updateUserMSIConfig.MaxRetry, "update-user-msi-max-retry", 2, "The maximum retry of UpdateUserMSI call")
-	flag.DurationVar(&updateUserMSIConfig.RetryPeriod, "update-user-msi-retry-period", 1*time.Second, "The duration to wait before retrying UpdateUserMSI")
+	flag.DurationVar(&updateUserMSIConfig.RetryInterval, "update-user-msi-retry-interval", 1*time.Second, "The duration to wait before retrying UpdateUserMSI")
 
 	flag.Parse()
 
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -44,19 +44,19 @@ const (
 	// have the correct role assignment to access a user-assigned identity.
 	linkedAuthorizationFailed retry.RetriableError = "LinkedAuthorizationFailed"
 	// Occurs when the user-assigned identity does not exist.
-	failedIdentityOperator retry.RetriableError = "FailedIdentityOperation"
+	failedIdentityOperation retry.RetriableError = "FailedIdentityOperation"
 )
 
 // NewCloudProvider returns a azure cloud provider client
-func NewCloudProvider(configFile string, updateUserMSIMaxRetry int, updateUseMSIRetryPeriod time.Duration) (c *Client, e error) {
+func NewCloudProvider(configFile string, updateUserMSIMaxRetry int, updateUseMSIRetryInterval time.Duration) (c *Client, e error) {
 	client := &Client{
 		configFile: configFile,
 	}
 	if err := client.Init(); err != nil {
 		return nil, err
 	}
-	client.RetryClient = retry.NewRetryClient(updateUserMSIMaxRetry, updateUseMSIRetryPeriod)
-	client.RetryClient.RegisterRetriableErrors(linkedAuthorizationFailed, failedIdentityOperator)
+	client.RetryClient = retry.NewRetryClient(updateUserMSIMaxRetry, updateUseMSIRetryInterval)
+	client.RetryClient.RegisterRetriableErrors(linkedAuthorizationFailed, failedIdentityOperation)
 	return client, nil
 }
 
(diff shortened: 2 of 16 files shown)


vickeyrihal1:
@chewong  thanks for working on this. Can we test it against multiple Identities? Maybe > 20, with 30% either non-existent or have permission issue?

cc: @dharmab

chewong:
@vickeyrihal1 you could use the mic image I built against this branch - `chewong/mic:retry-update-user-msi`. You might need to increase the `--update-user-msi-max-retry` value in MIC if you expect a large amount of invalid identities.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -37,14 +39,24 @@ type ClientInt interface {
 	Init() error
 }
 
+const (
+	// Occurs when the cluster service principal / managed identity does not
+	// have the correct role assignment to access a user-assigned identity.
+	linkedAuthorizationFailed retry.RetriableError = "LinkedAuthorizationFailed"
+	// Occurs when the user-assigned identity does not exist.
+	failedIdentityOperator retry.RetriableError = "FailedIdentityOperation"
```
aramase, line 47: nit: s/failedIdentityOperator/failedIdentityOperation

cmd/mic/main.go
```diff
@@ -81,6 +82,10 @@ func main() {
 	flag.StringVar(&typeUpgradeConfig.TypeUpgradeStatusKey, "type-upgrade-status-key", "type-upgrade-status", "Configmap key for type upgrade status")
 	flag.BoolVar(&typeUpgradeConfig.EnableTypeUpgrade, "enable-type-upgrade", true, "Enable type upgrade")
 
+	// Parameters for retrying cloudprovider's UpdateUserMSI function
+	flag.IntVar(&updateUserMSIConfig.MaxRetry, "update-user-msi-max-retry", 2, "The maximum retry of UpdateUserMSI call")
+	flag.DurationVar(&updateUserMSIConfig.RetryPeriod, "update-user-msi-retry-period", 1*time.Second, "The duration to wait before retrying UpdateUserMSI")
```
aramase, line 87: nit: `update-user-msi-retry-interval`

pkg/cloudprovider/vm.go
```diff
@@ -168,16 +169,39 @@ func (i *vmIdentityInfo) SetUserIdentities(ids map[string]bool) bool {
...
 
 	i.info.Type = getUpdatedResourceIdentityType(i.info.Type)
 	i.info.UserAssignedIdentities = userAssignedIdentities
 	return len(i.info.UserAssignedIdentities) > 0
 }
+
+func (i *vmIdentityInfo) RemoveUserIdentity(delID string) bool {
+	removed := false
+	if i.info.UserAssignedIdentities != nil {
+		for id := range i.info.UserAssignedIdentities {
```
aramase, line 184: `i.info.UserAssignedIdentities` is a map, so you can just check if the `delID` exists in the map instead of the for loop

pkg/cloudprovider/vm.go
```diff
@@ -168,16 +169,39 @@ func (i *vmIdentityInfo) SetUserIdentities(ids map[string]bool) bool {
...
+			if strings.EqualFold(id, delID) {
+				delete(i.info.UserAssignedIdentities, id)
+				removed = true
+				break
+			}
+		}
+	}
+
+	// Modify identity type in case we remove the last identity from the VM
+	if len(i.info.UserAssignedIdentities) == 0 {
```
aramase, line 194: `RemoveUserIdentity` is removing the identities from the desired list we are going to set (list which contains identities we want to assign and un-assign). So at the end of this, if `len(i.info.UserAssignedIdentities) == 0` then we should not be resetting the resource identity type. Length of 0 indicates we have no operation to perform, right?  But the identities which are still on the node should not be cleaned up.

pkg/cloudprovider/vmss.go
```diff
@@ -170,16 +171,39 @@ func (i *vmssIdentityInfo) SetUserIdentities(ids map[string]bool) bool {
...
 	i.info.Type = getUpdatedResourceIdentityType(i.info.Type)
 	i.info.UserAssignedIdentities = userAssignedIdentities
 	return len(i.info.UserAssignedIdentities) > 0
 }
+
+func (i *vmssIdentityInfo) RemoveUserIdentity(delID string) bool {
+	removed := false
+	if i.info.UserAssignedIdentities != nil {
+		for id := range i.info.UserAssignedIdentities {
+			if strings.EqualFold(id, delID) {
```
aramase, line 187: ditto as comment in vm

pkg/cloudprovider/vmss.go
```diff
@@ -170,16 +171,39 @@ func (i *vmssIdentityInfo) SetUserIdentities(ids map[string]bool) bool {
...
+				delete(i.info.UserAssignedIdentities, id)
+				removed = true
+				break
+			}
+		}
+	}
+
+	// Modify identity type in case we remove the last identity from the VMSS
+	if len(i.info.UserAssignedIdentities) == 0 {
+		i.ResetResourceIdentityType()
```
aramase, line 197: ditto

pkg/cloudprovider/vmss_test.go
```diff
@@ -22,3 +22,28 @@ func TestSetUserIdentities(t *testing.T) {
 	update = testIdentityInfo.SetUserIdentities(map[string]bool{"id3": true, "id4": true, "id1": false})
 	assert.True(t, update)
 }
+
+func TestRemoveUserIdentityVMSS(t *testing.T) {
+	testIdentityInfo := &vmssIdentityInfo{
+		info: &compute.VirtualMachineScaleSetIdentity{
+			UserAssignedIdentities: map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue{
+				"ID1": nil,
```
aramase, line 30: Lets change the value to `&compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue`

pkg/cloudprovider/vm_test.go
```diff
@@ -0,0 +1,49 @@
...
+	// add id3 and delete id1
+	update = testIdentityInfo.SetUserIdentities(map[string]bool{"id3": true, "id4": true, "id1": false})
+	assert.True(t, update)
+}
+
+func TestRemoveUserIdentityVM(t *testing.T) {
+	testIdentityInfo := &vmIdentityInfo{
+		info: &compute.VirtualMachineIdentity{
+			UserAssignedIdentities: map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue{
+				"ID1": nil,
```
aramase, line 30: Lets change the value to `&compute.VirtualMachineIdentityUserAssignedIdentitiesValue`

pkg/mic/mic.go
```diff
@@ -1099,14 +1107,27 @@ func (c *Client) updateUserMSI(newAssignedIDs map[string]aadpodid.AzureAssignedI
...
 
 			// Identity is successfully assigned to node, so update the status of assigned identity to assigned
 			if updateErr := c.updateAssignedIdentityStatus(&createID, aadpodid.AssignedIDAssigned); updateErr != nil {
 				message := fmt.Sprintf("Updating assigned identity %s status to %s for pod %s failed with error %v", createID.Name, aadpodid.AssignedIDAssigned, createID.Spec.Pod, updateErr.Error())
 				c.EventRecorder.Event(&createID, corev1.EventTypeWarning, "status update error", message)
 				klog.Error(message)
 			}
+
+			isCreateOperation := false
+			for _, i := range nodeTrackList.assignedIDsToCreate {
```
aramase, line 1120: This loop might be expensive when there are large # of identities to check

pkg/cloudprovider/vm.go
chewong: Makes sense. Let me modify the logic.

pkg/mic/mic.go
chewong: Will address this in a separate PR.

pkg/cloudprovider/vmss_test.go
chewong: Changed to `{}` since gofmt is complaining about `redundant type from array, slice, or map composite literalsimplifycompositelit` when I used `&compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue`.

Issue #618, opened 2020-05-15
Demo fails with failed list all vm azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request


bq1756:
**Describe the bug**
I followed the Demo instructions without modification.  Every command deployed what I expected or returned data I expected.  However, the demo pod log shows:

time="2020-05-15T16:28:18Z" level=info msg="starting demo pod demo/demo 10.207.40.24"
time="2020-05-15T16:29:38Z" level=error msg="failed list all vm azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to https://management.azure.com/subscriptions/2c1b98xxxxxxxx5f1b93f8/resourceGroups/xxRG_NAMExx/providers/Microsoft.Compute/virtualMachines?api-version=2017-12-01: StatusCode=403 -- Original Error: adal: Refresh request failed. Status Code = '403'. Response body: getting assigned identities for pod demo/demo in CREATED state failed after 16 attempts, retry duration [5]s. Error: <nil>\n" podip=10.207.40.24 podname=demo podnamespace=demo
time="2020-05-15T16:30:58Z" level=error msg="failed to refresh ServicePrincipalTokenFromMSI using the MSI VM extension, msiEndpoint(http://169.254.169.254/metadata/identity/oauth2/token)" podip=10.207.40.24 podname=demo podnamespace=demo

**Steps To Reproduce**
export SUBSCRIPTION_ID="2c1b98xxxxxxxx1b93f8"
export RESOURCE_GROUP="xxRG_NAMExx"
export IDENTITY_NAME="demo"
export CLUSTER_RG="MC_xxRG_NAMExx_centralus"
export VMSS="aks-agentpool1-35674147-vmss"
export AKS_CLUSTER_NAME="sa-aks2-nonprod"
export PRINCIPALID="360073xxxxxxxx73c85d"

az role assignment create --role "Managed Identity Operator" --assignee $PRINCIPALID --scope /subscriptions/$SUBSCRIPTION_ID/resourcegroups/$CLUSTER_RG
  - Returned successful JSON

az role assignment create --role "Virtual Machine Contributor" --assignee $PRINCIPALID --scope /subscriptions/$SUBSCRIPTION_ID/resourcegroups/$CLUSTER_RG
  - Returned successful JSON

kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml
[...]

PR #559, opened 2020-04-10
fix updating assigned identity when azure identity updated

aramase:
**Reason for Change**:

- Adds code path to update azure assigned identity when azure identity is updated in place
- Ensures old identity is deleted and new updated identity is assigned
- Adds e2e test to test update path

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/510, fixes https://github.com/Azure/aad-pod-identity/issues/504

**Notes for Reviewers**:

PR #620, opened 2020-05-15
docs: reduce ambiguity in demo and role assignment docs

chewong:
**Reason for Change**:

**Issue Fixed**:

Fixes https://github.com/Azure/aad-pod-identity/issues/618.

cc @bq1756

**Notes for Reviewers**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ export RESOURCE_GROUP="<ResourceGroup>"
 export IDENTITY_NAME="demo"
 ```
 
+> For AKS cluster, there are two resource groups that you need to be aware of - the resource group that contains the AKS cluster itself, and the cluster resource group (`MC_<AKSClusterName>_<AKSResourceGroup>_<Location>`). The latter contains all of the infrastructure resources associated with the cluster like VM/VMSS and VNet. Depending on where you deploy your user-assigned identities, you might need additional role assignments. Please refer to [Role Assignment](#role-assignment) for more information. For this demo, it is recommended to use the cluster resource group (the one with `MC_` prefix) as the `RESOURCE_GROUP` environment variable.
+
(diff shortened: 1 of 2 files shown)


elsesiy:
@chewong I think this PR should include an urgently needed reference to the need for `Network Contributor` when using Advanced Networking.

https://docs.microsoft.com/en-us/azure/aks/kubernetes-service-principal#networking

elsesiy:
Here's what I have assigned on a newly provisioned cluster with advanced networking and managed identities to make everything work for both the `Kubelet Identity` (e.g. k8s-agentpool as well as Cluster Identity (cat /etc/kubernetes/azure.json):
```
Managed Identity Operator (MC_RG)
Virtual Machine Contributor (MC_RG)
Network Contributor (Deployment RG)
```

chewong:
@elsesiy With https://github.com/Azure/aad-pod-identity/pull/559, I don't think the `Network Contributor` role is necessary anymore. I also tested with an AKS cluster with its VNet in a different resource group and aad-pod-identity works without assigning that particular role.

elsesiy:
@chewong I looked at the diff but don't see why this would solve this just yet. You'll run into this issue in multiple other occasions such as deploying an internal ALB as well so I needed it for sure.

aramase:
@elsesiy This is the [PR](https://github.com/Azure/aad-pod-identity/pull/522) that switches to using `PATCH` instead of VM `CreateOrUpdate`. `PATCH` only updates the identity config in the VM/VMSS. Are you using v1.6.0? What is the error you're seeing in MIC?

elsesiy:
@aramase I updated to `1.6.0` yesterday and haven't been seeing any errors since except for when deploying an internal load balancer. This case is unrelated to aad-pod-identity so I assume we're good here

PR #638, opened 2020-06-03
Add ability to pull ClientId and ResourceId from Secrets [WIP]

### LINKED ISSUE #532: MSI bindings - get client and resource ID from secrets
Hi, i am wondering if there will be a way to reference client and resource ID inside an AzureIdentity from Kubernetes Secrets?

Much like the example aadpodidentity-sp.yaml for using Service Principals as seen below

Service Principal Example
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: demo-aad1-sp
type: Opaque
data:
  ClientSecret: CLIENT_SECRET
---
kind: AzureIdentity
metadata:
  name: demo-aad1
spec:
  type: 1
  TenantID: TENANT_ID
  ClientID: CLIENT_ID
  ClientPassword: {"Name":"demo-aad1-sp","Namespace":"default"}
```

Idea for referencing values from secrets
```yaml
kind: AzureIdentity
metadata:
  name: demo-aad1
spec:
  type: 0
  ResourceID: {"Name":"demo-aad1-resource","Namespace":"product"}
  ClientID: {"Name":"demo-aad1-client","Namespace":"product"}
```

cwoolum:
**Reason for Change**:

This PR allows the ClientId and ResourceId to be pulled from secrets. This allows GitOps flows to work better so that infrastructure can be efficiently deployed anywhere without needing to individually update the `AzureIdentity` configuration.

**Issue Fixed**:

This fixes #532 

**Notes for Reviewers**:
This is currently a WIP. Some of the existing tests are broken because of how they compare the converted identity and the internal assigned identity. I'd like to get some feedback on the best way to update the tests so that the correct things are getting tested.

I thought it would make the most sense to put the secret fetching in the conversion helper because that way the internal azure identity would not need to care about secrets. I wasn't super happy about having the add `kubernetes.Interface` to the method signature of the methods though because it is only indirectly needed in some cases. I added tests to ensure that secrets were getting fetched correctly.

I have not done much Go development so would definitely value feedback on if this could be organized/implemented better!

### DIFF
--- a/pkg/apis/aadpodidentity/v1/conversion.go
+++ b/pkg/apis/aadpodidentity/v1/conversion.go
@@ -1,11 +1,16 @@
 package v1
 
 import (
+	"fmt"
 	"reflect"
 
 	aadpodid "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity"
 
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 func ConvertV1BindingToInternalBinding(identityBinding AzureIdentityBinding) (resIdentityBinding aadpodid.AzureIdentityBinding) {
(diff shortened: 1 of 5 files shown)


chewong:
@cwoolum thanks for opening this PR. Is this still WIP?

cwoolum:
@chewong, it's WIP because I need some feedback on how to update the existing tests that are broken. I think the new feature itself is complete though.

chewong:
Closing this PR for now. Please see https://github.com/Azure/aad-pod-identity/issues/532#issuecomment-651995514.

PR #643, opened 2020-06-09
Adjust CRDs in the Helm chart to v1 specification and be backward compatible

mstrzele:
**Reason for Change**:

It allows using the Helm chart on Kubernetes clusters in version 1.16 and above.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/charts/aad-pod-identity/Chart.yaml
+++ b/charts/aad-pod-identity/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Deploy components for aad-pod-identity
 name: aad-pod-identity
-version: 2.0.0
+version: 2.0.1
 appVersion: 1.6.1
 home: https://github.com/Azure/aad-pod-identity
 sources:
--- a/charts/aad-pod-identity/crds/crd.yaml
+++ b/charts/aad-pod-identity/crds/crd.yaml
@@ -23,6 +23,9 @@ spec:
   - name: v1
     served: true
     storage: true
+    schema:
+      openAPIV3Schema:
+        type: object
   {{- end }}
   names:
     kind: AzureAssignedIdentity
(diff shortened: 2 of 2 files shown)


mstrzele:
It seems that CRDs under the `crds` directory aren't templated. It will be hard or impossible to provide backward-compatible definitions to support Kubernetes 1.16+ deployments.

@chewong, what's your approach for this? Are you going to drop the support for older versions and update CRDs for both YAML manifests-based and Helm chart-based deployments?

mstrzele:
> It seems that CRDs under the `crds` directory aren't templated. It will be hard or impossible to provide backward-compatible definitions to support Kubernetes 1.16+ deployments.

> There is now a special directory called `crds` that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart.

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

chewong:
We are probably not going to drop support for `apiextensions.k8s.io/v1beta1` so older clusters can still use aad-pod-identity. I understand that `apiextensions.k8s.io/v1` is GA in 1.16 but as you mentioned, since templating is not available in the crd folder, I would continue using `apiextensions.k8s.io/v1beta1` for now.

aramase:
The removal of `v1beta1` is planned for 1.22 - https://github.com/kubernetes/kubernetes/issues/82022#issuecomment-636872808. Since helm doesn't support templating in crds dir, we can continue to use `apiextensions.k8s.io/v1beta1` to continue to support < 1.16 clusters. We will be able to switch to `apiextensions.k8s.io/v1` closer to 1.22 release and also make the default k8s version supported with helm chart to be 1.16+.

chewong:
This issue is being tracked in https://github.com/Azure/aad-pod-identity/issues/661. Closing this PR for now.

Issue #586, opened 2020-04-28
Move e2e tests to use client-go and azure sdk


aramase:
Currently e2e suite uses `kubectl apply -f` and `az vmss` commands for all tests. 

Long term plan - switch to using client-go and azure-sdk-go

Goals:
- Reduce test duration and technical debt by using various libraries instead of bash commands and Go-template-based deployments
- Make the e2e test suite easier to develop and run by following a similar approach to how Kubernetes and Kubernetes-related projects write their e2e framework and test cases. This will encourage more developers to write e2e tests when they open PRs.
- Enhance log dump / collection so we can debug test failures easier

PR #656, opened 2020-06-17
test: convert e2e test cases from old to new framework (part 2)

chewong:
**Reason for Change**:

Follow-up PR for #650.

Converted the following test cases from old to new framework:
`should assign the valid identity to the node when batch assigning identities that contain invalid identities`
`should not unassign existing identity if we assign invalid identities`
`should pass identity validation with correct identity and fail with wrong identity`
`should update AzureAssignedIdentity and when AzureIdentity fields are updated`
`should assign identity with init container`

**Issue Fixed**:

ref: #586, #657

**Notes for Reviewers**:

### DIFF
--- a/test/e2e_new/single_identity_test.go
+++ b/test/e2e_new/single_identity_test.go
@@ -85,6 +85,7 @@ var _ = Describe("[PR] When deploying one identity", func() {
 
 	It("should pass the identity validation", func() {
 		identityvalidator.Validate(identityvalidator.ValidateInput{
+			Getter:           kubeClient,
 			Config:           config,
 			KubeconfigPath:   kubeconfigPath,
 			PodName:          identityValidator.Name,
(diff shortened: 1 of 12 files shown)


test/e2e_new/single_identity_test.go
```diff
@@ -147,23 +150,79 @@ var _ = Describe("[PR] When deploying one identity", func() {
...
+		azureIdentity = azureidentity.Update(azureidentity.UpdateInput{
+			Updater:             kubeClient,
+			Config:              config,
+			AzureClient:         azureClient,
+			AzureIdentity:       azureIdentity,
+			UpdatedIdentityName: fmt.Sprintf("%s-0", keyvaultIdentity),
+		})
 
-	It("should pass liveness probe test", func() {
+		azureassignedidentity.WaitForLen(azureassignedidentity.WaitForLenInput{
```
aramase, line 162: The `AzureAssignedIdentity` is already 1 at this point. So when we validate it's possible MIC is actually assigning the new identity and we'll get an error `"identity not found"`. The status is flipped to `Created` right before the update, so we should also check if the `AzureAssignedIdentity` state is `Assigned` before validating

Issue #619, opened 2020-05-15
Change NMI to listen only on localhost


aramase:
Currently NMI binds on all the interfaces. Need to change that to only listen on `localhost:2579`

PR #658, opened 2020-06-17
feat: make NMI listen only on localhost

aramase:
**Reason for Change**:

By default the http server binds on all interfaces. This changes to only listen on localhost as it's secure.

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/619

**Notes for Reviewers**:


aramase:
Removed `--host-ip` flag in this PR. Opened https://github.com/Azure/aad-pod-identity/issues/659 to track removing flag and reference from deployments as part of next release.

chewong:
Will upgrade be affected by the removal of the flag?

aramase:
> Will upgrade be affected by the removal of the flag?

No, it should not be. It's a rolling upgrade, so old NMI is deleted, cleans up iptable rules as it gracefully terminates. Then new scheduled NMI pods should add the new iptable rules with `localhost`. Upgrading through `helm` or `kubectl apply` will not result in error if the arg is removed.

chewong:
Hmm not sure why the CI test pass w/ new test framework is flaky. I will re-run the tests in #656 once this is merged.

aramase:
Most of the timeouts seem to be around namespace deletion. I would assume `AzureAssignedIdentity` is lingering around and blocking ns deletion, but we are not running in forced namespaced mode, so that doesn't explain it as all `AzureAssignedIdentities` are in default ns.

chewong:
> Most of the timeouts seem to be around namespace deletion. I would assume `AzureAssignedIdentity` is lingering around and blocking ns deletion, but we are not running in forced namespaced mode, so that doesn't explain it as all `AzureAssignedIdentities` are in default ns.

```
  <*errors.StatusError>: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"azureassignedidentities.aadpodidentity.k8s.io \"identity-validator-kj9rn-multiple-identities-hm6pp-keyvault-identity-3\" not found", Reason:"NotFound", Details:(*v1.StatusDetails)(0xc0003edb00), Code:404}}
```

In `should create AzureAssignedIdentities for 40 pods within 150 seconds` test case, I think we should wait for all 40 AzureAssignedIdentities to be created before validating the pod. Anyway, merging this PR for now and I will fix it in #656.

Issue #657, opened 2020-06-17
test: convert E2E test cases from old to new test framework


chewong:
| Test Case                                                                                                                            | Status              |
|--------------------------------------------------------------------------------------------------------------------------------------|---------------------|
| `should pass the identity validating test`                                                                                           | Migrated in #650    |
| `should not pass the identity validating test if the AzureIdentity is deleted`                                                       | Migrated in #650    |
| `should not pass the identity validating test if the AzureIdentityBinding is deleted`                                                | Migrated in #650    |
| `should delete the AzureAssignedIdentity if the deployment is deleted`                                                               | Migrated in #650    |
| `should establish a new AzureAssignedIdentity and remove the old one when draining the node containing identity validator`           | Migrating in #664   |
| `should remove the correct identities when adding AzureIdentity and AzureIdentityBinding in order and removing them in random order` | Migrated in #650    |
| `should not alter the user assigned identity on VM after AAD pod identity is created and deleted`                                    | Migrated in #662    |
| `should cleanup iptable rules after deleting aad-pod-identity`                                                                       | Migrated in #650    |
| `should not alter the system assigned identity after creating and deleting pod identity`                                             | Migrated in #662    |
| `should create azureassignedidentities for 40 pods within ~2mins 30seconds`                                                          | Migrated in #650    |
| `should be backward compatible with old and new version of mic and nmi`                                                              | Migrating in #664   |
| `should not delete an in use identity from a vmss`                                                                                   | Migrated in #662    |
| `should not delete the Immutable Identity from vmss when the deployment is deleted`                                                  | Migrating in Part 5 |
| `should pass liveness probe test`                                                                                                    | Migrating in #664   |
| `should pass validation by bypassing nmi using azurepodidentityexception crd`                                                        | Migrating in Part 5 |
| `should pass identity validation with correct identity and fail with wrong identity`                                                 | Migrated in #656    |
| `should delete assigned identity when identity no longer exists on underlying node`                                                  | Migrated in #662    |
| `should pass multiple identity validating test even when MIC is failing over`                                                        | Migrating in #664   |
| `should assign identity with init containers`                                                                                        | Migrated in #656    |
| `should pass the identity format validation with gatekeeper constraint`                                                              | Migrated in Part 5  |
| `should assign new identity and remove old when AzureIdentity updated`                                                               | Migrated in #656    |
| `should assign the valid identity to the node when batch assigning identities that contain invalid identities`                       | Migrated in #656    |
[...]

PR #662, opened 2020-06-22
test: convert e2e test cases from old to new framework (part 3)

chewong:
**Reason for Change**:

Follow-up PR for #656.

Converted the following test cases from old to new framework:
`should not delete a user-assigned identity that is being used by a different pod`
`should be able to delete AzureAssignedIdentity when the user-assigned is un-assigned from the underlying node`
`should not alter the system-assigned identity after creating and deleting pod identity`
`should not alter the user-assigned identity on VM after AAD pod identity is created and deleted`

**Issue Fixed**:

ref: #586, #657

**Notes for Reviewers**:


test/e2e_new/node_test.go
```diff
@@ -3,23 +3,311 @@
...
+		})
 	})
 
-	It("should not alter the system assigned identity after creating and deleting pod identity", func() {
+	It("should not delete a user-assigned identity that is being used by a different pod", func() {
+		// This test is specifically testing vmss behavior
+		// As such we'll look through the cluster to see if there are nodes assigned
+		// to a vmss, and if any of thoe vmss's have more than one node.
+		//
+		// We cannot do the test if there is not at least1 vmss with at least 2 nodes
```
aramase, line 79: nit: s/least1/least 1

PR #663, opened 2020-06-23
fix: check iptable rules match expected

aramase:
**Reason for Change**:

Checking if the length of the iptable rules in the `aad-metadata` chain equals 3 is not sufficient as it's possible the rules were changed underneath. To ensure the rules are what we expect it to be, check if existing rules match all expected rules. If any difference in the rules, then flush and recreate the chain.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/nmi/iptables/iptables.go
+++ b/pkg/nmi/iptables/iptables.go
@@ -84,14 +84,33 @@ func ensureCustomChain(ipt *iptables.IPTables, destIP, destPort, targetip, targe
 	/*
 		iptables -t nat -S aad-metadata returns 3 rules
 			-N aad-metadata
-			-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 10.240.3.134:2579
+			-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:<nmi port>
 			-A aad-metadata -j RETURN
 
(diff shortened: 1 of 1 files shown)


chewong:
You could ignore the CI errors from the new framework. I included improvements in #656 & #662 which should reduce the error rate.

aramase:
sounds good! Are the 2 PRs ready for review?

chewong:
Both are ready for review but #656 should get merged first since #662 depends on it

pkg/nmi/iptables/iptables.go
```diff
@@ -89,9 +89,28 @@ func ensureCustomChain(ipt *iptables.IPTables, destIP, destPort, targetip, targe
 
 		For this reason we check if the length of rules is 3. If not 3, then we flush and create chain again.
 	*/
-	if len(rules) == 3 {
+
```
chewong, line 92: nit: `s/10.240.3.134:2579/127.0.0.1:<nmi port>` in the comment block above.

pkg/nmi/iptables/iptables.go
```diff
@@ -89,9 +89,28 @@ func ensureCustomChain(ipt *iptables.IPTables, destIP, destPort, targetip, targe
...
+	expectedRules := map[string]struct{}{
+		"-N aad-metadata": {},
+		"-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:2579": {},
```
chewong, line 95: `2579` is the default NMI port but it can be changed via `--nmi-port` flag. Let's account that scenario as well.

PR #668, opened 2020-06-30
add default known types to scheme

aramase:
**Reason for Change**:

- Adds default known types to scheme to prevent errors while reestablishing watch due to network timeouts

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers/internalinterfaces"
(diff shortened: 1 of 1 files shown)


aramase:
> What's the benefit of using NewSchemeBuilder over s := runtime.NewScheme() + s.AddKnownTypes? 

Schemebuilder is used to collect functions to add to scheme and that's mostly how `controller-runtime` does it now for custom resource definitions. But I guess we can stick to just adding to scheme for now as it's just a single scheme on top of the default known types.

aramase:
> What's the benefit of using NewSchemeBuilder over s := runtime.NewScheme() + s.AddKnownTypes? 

Schemebuilder is used to collect functions to add to scheme and that's mostly how `controller-runtime` does it now for custom resource definitions. But I guess we can stick to just adding to scheme for now as it's just a single scheme on top of the default known types.

PR #670, opened 2020-07-01
Remove unused cert volumes from mic deployment

omegas27:
**Reason for Change**:

Remove unused /etc/kubernetes/certs directory as this is sensitive information


aramase:
@chewong For the `deploy` folder, let's remove it in the next release. Let's not change the yamls now because users could be using that. We should effectively tie this change to a release. WDYT?

chewong:
SGTM

> @chewong For the `deploy` folder, let's remove it in the next release. Let's not change the yamls now because users could be using that. We should effectively tie this change to a release. WDYT?

aramase:
Opened issue to track this cleanup in `deploy` dir - https://github.com/Azure/aad-pod-identity/issues/672

aramase:
@chewong Going to merge this if you have no other comments?

PR #687, opened 2020-07-14
ci: add soak testing as part of nightly build & test and remove Jenkinsfile

chewong:
**Reason for Change**:

**Issue Fixed**:

fixes #654.

**Notes for Reviewers**:

PR #693, opened 2020-07-20
fix: check if identity exists before assign in e2e

aramase:
**Reason for Change**:

The case of the `resourceID` returned from the `GET` can differ from the `resourceID` we construct. It's best to check if the identity exists, and if does skip the assignment. This should resolve the flakes we are seeing with assigning identity during the e2e runs.

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/framework/azure/vm_manager.go
+++ b/test/e2e/framework/azure/vm_manager.go
@@ -53,7 +53,15 @@ func (m *vmManager) AssignUserAssignedIdentity(vmName, identityToAssign string) 
 		}
 	}
 
-	vm.Identity.UserAssignedIdentities[fmt.Sprintf(ResourceIDTemplate, m.config.SubscriptionID, m.config.IdentityResourceGroup, identityToAssign)] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
+	identityAssignResourceID := fmt.Sprintf(ResourceIDTemplate, m.config.SubscriptionID, m.config.IdentityResourceGroup, identityToAssign)
+	for identity := range vm.Identity.UserAssignedIdentities {
+		// identity already exists and doesn't need to be re-assigned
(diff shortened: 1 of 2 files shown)


test/e2e/framework/azure/vm_manager.go
```diff
@@ -53,7 +53,15 @@ func (m *vmManager) AssignUserAssignedIdentity(vmName, identityToAssign string)
...
-	vm.Identity.UserAssignedIdentities[fmt.Sprintf(ResourceIDTemplate, m.config.SubscriptionID, m.config.IdentityResourceGroup, identityToAssign)] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
+	identityAssignResourceID := fmt.Sprintf(ResourceIDTemplate, m.config.SubscriptionID, m.config.IdentityResourceGroup, identityToAssign)
+	for identity := range vm.Identity.UserAssignedIdentities {
+		// identity already exists and doesn't need to be re-assigned
+		if strings.EqualFold(identity, identityAssignResourceID) {
+			return nil
+		}
+	}
+
+	vm.Identity.UserAssignedIdentities[identityAssignResourceID] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
```
chewong, line 64: Got the following error https://dev.azure.com/azure/aad-pod-identity/_build/results?buildId=12327&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=c0307c8e-cb29-5919-a192-498b9fd8bfd8&l=217 in #687. Do you know if it's possible to have `vm.Identity`not equal to nil but `vm.Identity.UserAssignedIdentities` is nil?

aramase: It is possible if the SDK is setting to nil. In our actual code path, we do check if its nil and initialize if that's the case. Let me update this PR to do that.

PR #697, opened 2020-07-21
fix: set max pods and assign empty map for system-assigned identity cleanup

aramase:
**Reason for Change**:

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/identity_exception_test.go
+++ b/test/e2e/identity_exception_test.go
@@ -50,8 +50,12 @@ var _ = Describe("[PR] When deploying AzurePodIdentityException", func() {
 
 		err := azureClient.AssignUserAssignedIdentity(node.Spec.ProviderID, keyvaultIdentity)
 		Expect(err).To(BeNil())
+		err = azureClient.EnableSystemAssignedIdentity(node.Spec.ProviderID)
+		Expect(err).To(BeNil())
 		defer func() {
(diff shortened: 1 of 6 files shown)


test/e2e/identity_exception_test.go
```diff
@@ -50,9 +50,13 @@ var _ = Describe("[PR] When deploying AzurePodIdentityException", func() {
 
 		err := azureClient.AssignUserAssignedIdentity(node.Spec.ProviderID, keyvaultIdentity)
 		Expect(err).To(BeNil())
+		err = azureClient.EnableSystemAssignedIdentity(node.Spec.ProviderID)
```
chewong, line 53: Why are we enabling system-assigned identity here?

aramase: With identity validator there are 2 tests performed -
1. Testing getting a token for User MSI (passing the clientID or resourceID)
2. Testing getting a token for System MSI 
  - By default `aad-pod-identity` will fetch the token the first clientID of User MSI that matches the pod.
  - In case the pod is excepted, the request is always forwarded as is to the IMDS. For that scenario a `system-assigned-identity` needs to exist or the test will fail.

aramase: Also, this issue never showed up in the previous runs because the system assigned identity was never cleaned up after the node tests are run. If this test runs before the node test, then we would've seen a flake.

chewong: Makes sense

PR #705, opened 2020-07-23
ci: add code coverage as part of CI

### LINKED ISSUE #704: ci: add code coverage tool

chewong:
**Reason for Change**:

**Issue Fixed**:

fixes #704

**Notes for Reviewers**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -323,3 +323,6 @@ __pycache__/
 
 # Go tools
 .tools/
+
+# cover profile
+coverage.txt
--- a/.pipelines/unit-tests-template.yml
+++ b/.pipelines/unit-tests-template.yml
@@ -12,6 +12,10 @@ jobs:
         displayName: Build
       - script: make unit-test
         displayName: Run unit tests
+      - script: bash <(curl -s https://codecov.io/bash)
+        displayName: Upload coverage to codecov
+        env:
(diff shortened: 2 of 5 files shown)


.pipelines/e2e-tests-template.yml
```diff
@@ -15,7 +15,6 @@ jobs:
         workspace:
           clean: all
         variables:
-          - group: aad-pod-identity
```
aramase, line 18: Won't moving this affect `e2e-tests` for nightly job?

.pipelines/pr.yml
```diff
@@ -9,7 +9,10 @@ pr:
       - docs/*
 
 pool:
-  vmImage: ubuntu-latest
+  vmImage: ubuntu-18.04
```
aramase, line 12: What is the reason for not using the latest?

.pipelines/e2e-tests-template.yml
chewong: Right, originally I put the CODECOV_TOKEN within the `aad-pod-identity` variable group but since I will set it as a secret variable in the PR pipeline.

.pipelines/pr.yml
chewong: I wanted to specify the image version instead of using the latest, which could be changed in the future. But I doubt it would affect our test result. I can change it back to latest

Issue #677, opened 2020-07-08
Honor retry-after header or internal backoff on 429


aramase:
**Describe the request**

**Explain why AAD Pod Identity needs it**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

mhsh64:
@aramase hi, do we know if this is something can be fixed in your next release?
We randomly getting failure during the AKS upgrade

aramase:
@mhsh64 What is the error you're getting? We'll plan this for next release. Do you know what in the cluster is causing the throttling?

mhsh64:
Hi @aramase 
This related to #676 
MSFT support advised me to bring down all MICs (0) and upgrading, which it worked. But for production it may not possible to do that

aramase:
Thank you for the details, @mhsh64. We'll plan this for next release.

mhsh64:
@aramase 
Hi
We still see throttling causing by MIC

![21BD81E8-3AC4-4AC0-82D4-34519D42C6F6](https://user-images.githubusercontent.com/14878974/104963173-ecab6c80-59a7-11eb-9528-efe90e9125eb.jpeg)


Do you have any idea?

aramase:
@mhsh64 This PR adds support to honor retry after header that is sent back for the API request. This however doesn't completely prevent throttling from happening. If multiple components are making ARM API calls, then the throttling limit will be hit. With this change MIC detects the API request is throttled and instead of repeatedly retrying it'll only retry the request at a time t + retryAfter as recommended by the ARM response.

PR #742, opened 2020-08-17
feat: throttling - honor retry after header

aramase:
**Reason for Change**:

- Exclude `429` from retriable error codes to prevent SDK from making retries.
- Check the response in future/GET result to check if `Retry-After` header exists and cache the back off value in client.

**Requirements**

- [x] squashed commits
- [ ] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

fixes https://github.com/Azure/aad-pod-identity/issues/677

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -3,9 +3,11 @@ package cloudprovider
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -45,6 +47,12 @@ const (
 	linkedAuthorizationFailed retry.RetriableError = "LinkedAuthorizationFailed"
 	// Occurs when the user-assigned identity does not exist.
 	failedIdentityOperation retry.RetriableError = "FailedIdentityOperation"
+	// retryAfterHeaderKey is the retry-after header key in ARM responses.
+	retryAfterHeaderKey = "Retry-After"
+)
+
+var (
+	now = time.Now
 )
 
 // NewCloudProvider returns a azure cloud provider client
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -45,6 +47,12 @@ const (
 	linkedAuthorizationFailed retry.RetriableError = "LinkedAuthorizationFailed"
 	// Occurs when the user-assigned identity does not exist.
 	failedIdentityOperation retry.RetriableError = "FailedIdentityOperation"
+	// retryAfterHeaderKey is the retry-after header key in ARM responses.
+	retryAfterHeaderKey = "Retry-After"
+)
+
+var (
+	now = time.Now
```
chewong, line 55: Why are we re-defining the `time.Now` function as a separate variable? Let's standardize how we grab the current time in the package by using one of `now()` and `time.Now()`.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -329,3 +347,32 @@ func extractIdentitiesFromError(err error) []string {
...
+	}
+
+	var dur time.Duration
+	if retryAfter, _ := strconv.Atoi(ra); retryAfter > 0 {
+		dur = time.Duration(retryAfter) * time.Second
+	} else if t, err := time.Parse(time.RFC1123, ra); err == nil {
+		dur = t.Sub(now())
+	}
+	return dur
+}
```
chewong, line 378: Let's write some unit tests for this function.

pkg/cloudprovider/vm.go
```diff
@@ -77,8 +80,19 @@ func (c *VMClient) Get(rgName string, nodeName string) (compute.VirtualMachine,
...
+	// Report errors if the client is throttled.
+	if c.retryAfterReader.After(time.Now()) {
+		return compute.VirtualMachine{}, fmt.Errorf("VMGet client throttled, retry after: %v", c.retryAfterReader)
+	}
+
 	vm, err := c.client.Get(ctx, rgName, nodeName, "")
 	if err != nil {
+		resp := vm.Response.Response
+		retryAfterDuration := getRetryAfter(resp)
+		if isThrottled(resp) || retryAfterDuration != 0 {
```
chewong, line 92: We can probably merge `isThrottled` with `getRetryAfter` since every 429 requests should have a `Retry-After` header. So if a request is not being throttled, `getRetryAfter` will just return 0.

pkg/cloudprovider/vm.go
```diff
@@ -22,6 +22,9 @@ import (
 type VMClient struct {
 	client   compute.VirtualMachinesClient
 	reporter *metrics.Reporter
+	// ARM throttling configures.
+	retryAfterReader time.Time
+	retryAfterWriter time.Time
```
chewong, line 27: Are there going to be any race conditions for these two variables? If there are two 429 responses, the one that is being processed later will overwrite the `retryAfter*` time set by the first request.

aramase: The `retryAfter` is per client which means we always want the latest possible retry after time. So if it gets overwritten to a later time by the second 429 response, then that is what we want to use.

PR #785, opened 2020-09-08
chore: add logs for ignored pods

### LINKED ISSUE #732: MIC only sporadically assigning identities to pods
This is a weird issue, so its hard to even figure out where something might be failing. We've been testing aad-pod-identity as part of migrating service hosting for one of our products from Azure ASE to Kubernetes for a few weeks now. The product relies on MSI in ASE, and we needed aad-pod-identity to allow us to access Azure services without having to refactor code, as well as to pull secrets from KeyVault. 

Its been working fine for a couple weeks, but starting a couple of days ago, all of our test clusters -- some of which hadn't been updated since early last week -- had MIC fail on us when we pushed out a deployment that caused pods to be recycled. The new pods in the existing cluster couldn't retrieve identities -- NMI would return a 404 error. 

MIC was basically not picking up any changes -- it wasn't responding to events at all. 

We've deleted the resource groups and clusters a few times in the last couple of days trying to figure out what is going on, and the problem seems to be consistent. Our clusters all are created from scripts, so the test clusters today are identically configured to the ones a week ago.  What we've found is that we (generally) get identities assigned correctly on a brand new cluster. But after that happens, we never do again. There's no MIC logging showing up at all when the pod states change.

We just get this:
```
I0811 16:01:14.110326       1 crd.go:453] CRD informers started
I0811 16:01:14.510233       1 pod.go:72] pod cache synchronized. Took 500.20951ms
I0811 16:01:14.510257       1 pod.go:79] pod watcher started !!
I0811 16:01:14.610290       1 mic.go:390] sync thread started.
```

Doing a restart of pods, nothing shows up. Doing a reconfigure, same thing. Nothing shows up.  

If we then delete all the deployments, again nothing shows up. Now, strangely, when we then try re-installing the deployments, it wakes up and attempts to delete the identities assigned to the deployments that were already deleted. But it never adds them to the new ones. In some cases, where the system was cycling through pod deployments rapidly as configurations are getting updated, it'll get further behind. Like logs "now" on an update will have an identity being removed from a pod that has been gone for a few updates. And nothing else.

Restarting MIC via a rollout / restart permanently breaks it -- it no longer responds to anything. Deleting and recreating the cluster from scratch is the only fix we've found. MIC never reports any errors -- when it first works, there are no errors and subsequent to that when it is no longer responding to events, there are also no errors. 

NMI continuously returns the same error:

`E0811 16:35:21.068262       1 server.go:232] failed to get identities, error: getting assigned identities for pod default/apiserver-846fb5c54c-q6v59 in CREATED state failed after 16 attempts, retry duration [5]s, error: <nil>`

The clusters are MSI-enabled via --enable-managed-identity, the roles are all correct in the resource groups. (The identities are able to be set once, so that's clearly okay.) 

The aad-pod-identity components are installed via:

```
kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/deployment-rbac.yaml
kubectl apply -f https://raw.githubusercontent.com/Azure/aad-pod-identity/master/deploy/infra/mic-exception.yaml
```

The installed version, now and when it was reliably working, both report 1.6.2. The clusters have been Kubernetes 1.18.4.

I recognize this is a lousy bug report, but I'm scratching my head and not really sure what I can do to even get more information. The lack of any errors isn't giving me any clue why MIC just stops handling any cluster updates.

aramase:
**Reason for Change**:

- Adds logs for when pods are ignored because of missing labels, misconfigured label or no matching AzureIdentityBinding found for the label.

**Requirements**

- [x] squashed commits
- [ ] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

fixes #732 

**Notes for Reviewers**:

### DIFF
--- a/pkg/nmi/managed.go
+++ b/pkg/nmi/managed.go
@@ -84,12 +84,12 @@ func (mc *ManagedClient) GetToken(ctx context.Context, rqClientID, rqResource st
 	case aadpodid.ServicePrincipal:
 		tenantID := azureID.Spec.TenantID
 		adEndpoint := azureID.Spec.ADEndpoint
-		secretName := &azureID.Spec.ClientPassword
+		secretRef := &azureID.Spec.ClientPassword
 		klog.Infof("matched identityType:%v adendpoint:%s tenantid:%s clientid:%s resource:%s",
 			idType, adEndpoint, tenantID, utils.RedactClientID(clientID), rqResource)
-		secret, err := mc.KubeClient.GetSecret(secretName)
+		secret, err := mc.KubeClient.GetSecret(secretRef)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get Kubernetes secret %s, err: %v", secretName, err)
+			return nil, fmt.Errorf("failed to get Kubernetes secret %s/%s, err: %v", secretRef.Namespace, secretRef.Name, err)
 		}
 		clientSecret := ""
 		for _, v := range secret.Data {
@@ -101,11 +101,12 @@ func (mc *ManagedClient) GetToken(ctx context.Context, rqClientID, rqResource st
 	case aadpodid.ServicePrincipalCertificate:
 		tenantID := azureID.Spec.TenantID
 		adEndpoint := azureID.Spec.ADEndpoint
+		secretRef := &azureID.Spec.ClientPassword
 		klog.Infof("matched identityType:%v adendpoint:%s tenantid:%s clientid:%s resource:%s",
 			idType, adEndpoint, tenantID, utils.RedactClientID(clientID), rqResource)
-		secret, err := mc.KubeClient.GetSecret(&azureID.Spec.ClientPassword)
+		secret, err := mc.KubeClient.GetSecret(secretRef)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to get Kubernetes secret %s/%s, err: %v", secretRef.Namespace, secretRef.Name, err)
 		}
 		certificate, password := secret.Data["certificate"], secret.Data["password"]
 		token, err := auth.GetServicePrincipalTokenWithCertificate(adEndpoint, tenantID, clientID,
--- a/pkg/nmi/managed.go
+++ b/pkg/nmi/managed.go
@@ -84,11 +84,12 @@ func (mc *ManagedClient) GetToken(ctx context.Context, rqClientID, rqResource st
 	case aadpodid.ServicePrincipal:
 		tenantID := azureID.Spec.TenantID
 		adEndpoint := azureID.Spec.ADEndpoint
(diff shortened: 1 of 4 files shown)


pkg/nmi/managed.go
```diff
@@ -84,11 +84,12 @@ func (mc *ManagedClient) GetToken(ctx context.Context, rqClientID, rqResource st
...
 		tenantID := azureID.Spec.TenantID
 		adEndpoint := azureID.Spec.ADEndpoint
+		secretName := &azureID.Spec.ClientPassword
 		klog.Infof("matched identityType:%v adendpoint:%s tenantid:%s clientid:%s resource:%s",
 			idType, adEndpoint, tenantID, utils.RedactClientID(clientID), rqResource)
-		secret, err := mc.KubeClient.GetSecret(&azureID.Spec.ClientPassword)
+		secret, err := mc.KubeClient.GetSecret(secretName)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to get Kubernetes secret %s, err: %v", secretName, err)
```
chewong, line 92: Let's apply the same error message for `aadpodid.ServicePrincipalCertificate` here and `pkg/nmi/standard.go`

pkg/nmi/standard.go
```diff
@@ -181,11 +181,12 @@ func (sc *StandardClient) GetToken(ctx context.Context, rqClientID, rqResource s
 	case aadpodid.ServicePrincipal:
 		tenantID := azureID.Spec.TenantID
 		adEndpoint := azureID.Spec.ADEndpoint
+		secretName := &azureID.Spec.ClientPassword
```
chewong, line 184:
```suggestion
		secretRef := &azureID.Spec.ClientPassword
```

pkg/nmi/standard.go
```diff
@@ -181,11 +181,12 @@ func (sc *StandardClient) GetToken(ctx context.Context, rqClientID, rqResource s
...
 		klog.Infof("matched identityType:%v adendpoint:%s tenantid:%s clientid:%s resource:%s",
 			idType, adEndpoint, tenantID, utils.RedactClientID(clientID), rqResource)
-		secret, err := sc.KubeClient.GetSecret(&azureID.Spec.ClientPassword)
+		secret, err := sc.KubeClient.GetSecret(secretName)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to get Kubernetes secret %s, err: %v", secretName, err)
```
chewong, line 189:
```suggestion
		secret, err := sc.KubeClient.GetSecret(secretRef)
		if err != nil {
			return nil, err
			return nil, fmt.Errorf("failed to get Kubernetes secret %s/%s, err: %v", secretRef.Namespace, secretRef.Name, err)
```

chewong (approved):

PR #847, opened 2020-10-23
fix: account for 150+ identity assignment and unassignment

### LINKED ISSUE #813: MIC hitting (un-)assignment limit of 150
**Describe the bug**

Whenever MIC needs to either assign or unassign >150 (in one cycle?) - it fails with the following message:

```
...
User Assigned Identities Number '165' Assign exceed limit '150' in one request
...
```

This is the same error message as the issue I've raised previously: https://github.com/Azure/aad-pod-identity/issues/637 before. We were on v1.5.4 previously but we upgraded to v1.6.3 last night and it ran into the same issue.

When this happens, no new assignments would go through.

The only way to "unstuck" it was by manual intervention. We removed some role assignments from the VMSS via the UI, to lower the number of un-assignments that it was trying to do. 

**Steps To Reproduce**

Hypothetical steps that I think will reproduce it but I'm not 100% on how it arrived at this state atm:

- Stop MIC operations (e.g. scale it down first or something to simulate a CLBO)
- Manually assign 150 role assignments that are not needed onto the VMSS role assignment via the UI. These should be 
- Start MIC 
- I presume it would try and un-assign a list >150

**Expected behavior**

- If the number of assignment/un-assignments is greater than 150, it should potentially split it up into multiple cycles 

**AAD Pod Identity version**
v1.6.3

**Kubernetes version**
1.16.10

**Additional context**

This is on role assignments to a VMSS not to individual nodes.

chewong:

**Reason for Change**:

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

fixes #813

**Please answer the following questions with yes/no**: no

Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution? No

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/truncate.go
+++ b/pkg/cloudprovider/truncate.go
@@ -1,0 +1,39 @@
+package cloudprovider
+
+import (
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute"
+)
+
+const (
+	maxIdentitiesCount = 150
+)
+
+// truncateVMIdentities truncates a given list of vm identities to maxIdentitiesCount and returns any extra identities.
+func truncateVMIdentities(ids map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue) (map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue, map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue) {
+	rest := make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+	i := 0
+	for k, v := range ids {
+		if i >= maxIdentitiesCount {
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/truncate.go
```diff
@@ -0,0 +1,39 @@
...
+			rest[k] = v
+			delete(ids, k)
+		}
+		i++
+	}
+
+	return ids, rest
+}
+
+// truncateVMIdentities truncates a given list of vmss identities to maxIdentitiesCount and returns any extra identities.
```
aramase, line 26: nit: s/truncateVMIdentities/truncateVMSSIdentities

aramase (approved):
Just one small nit, but otherwise lgtm!

PR #842, opened 2020-10-19
chore: disable aad-pod-identity by default for kubenet

aramase:
**Reason for Change**:

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] added unit tests and e2e tests (if applicable). See [test standard](https://github.com/Azure/aad-pod-identity/blob/master/CONTRIBUTING.md#test-standard) for more details.
- [x] ran `make precommit`

**Issue Fixed**:

**Please answer the following questions with yes/no**:

Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?

Some of the docs contain snippets from https://blog.aquasec.com/dns-spoofing-kubernetes-clusters

**Notes for Reviewers**:

PR #888, opened 2020-12-02
Adding option to configure kubeletConfig

### LINKED ISSUE #887: Allow kubeletConfig to be configurable in NMI
**Describe the request**
Allow the kubeletConfig path to be configurable in the NMI DaemonSet.

For non-default AKS-Engine installations, this path may not be in `/etc/default/kubelet` which will cause a directory to be created in its current form.

Add value `.Values.nmi.kubeletConfig` and conditional check in nmi-daemonset to apply the flag.

1. Add `type: FileOrCreate` to the hostPath properties like `iptableslock` is configured
1. Symlink/copy kubelet config to `/etc/default/kubelet` on each Worker node.

**Additional context**

maxirus:
**Reason for Change**:

Adds ability to set the kubeletConfig path.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [ ] ran `make precommit`

**Issue Fixed**:

Fixes #887 

Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution? No

**Notes for Reviewers**:


chewong (changes requested):
Could you make the changes in `manifest_staging/charts`, which stages all the changes before a release?

chewong (commented):
```diff
diff --git a/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml b/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
index 383f123..1762aad 100644
--- a/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
+++ b/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
@@ -44,7 +44,7 @@ spec:
         name: iptableslock
       - name: kubelet-config
         hostPath:
-          path: /etc/default/kubelet
+          path: {{ .Values.nmi.kubeletConfig }}
       containers:
       - name: nmi
         image: "{{ .Values.image.repository }}/{{ .Values.nmi.image }}:{{ .Values.nmi.tag }}"
```

chewong (dismissed):

aramase (approved):

chewong:
> [CLA assistant check](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888) before we can accept your contribution.
> 
> ❌ maxirus [sign now](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888)
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.microsoft.com/check/Azure/aad-pod-identity?pullRequest=888) it.

@maxirus Could you also sign the CLA?

maxirus:
> > [CLA assistant check](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888)
> > Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888) before we can accept your contribution.
> > ❌ maxirus [sign now](https://cla.opensource.microsoft.com/Azure/aad-pod-identity?pullRequest=888)
> > You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.microsoft.com/check/Azure/aad-pod-identity?pullRequest=888) it.
> 
> @maxirus Could you also sign the CLA?

I tried but it wasn’t working.

maxirus:
I agree to CLA

chewong:
Looks like the backward compatibility test is failing because the older version does not have the `--kubelet-config` flag. 

Let's change the default value of `nmi.kubeletConfig` in `value.yaml` to an empty string (since the flag default is already set in https://github.com/Azure/aad-pod-identity/blob/master/cmd/nmi/main.go#L52) so the flag is not used unless it's specified by the user. For the path in the spec, we can default it to `/etc/default/kubelet`:

```diff
diff --git a/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml b/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
index b075928..5796069 100644
--- a/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
+++ b/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml
@@ -44,7 +44,7 @@ spec:
         name: iptableslock
       - name: kubelet-config
         hostPath:
-          path: {{ .Values.nmi.kubeletConfig }}
+          path: {{ .Values.nmi.kubeletConfig | default "/etc/default/kubelet" }}
       containers:
       - name: nmi
         image: "{{ .Values.image.repository }}/{{ .Values.nmi.image }}:{{ .Values.nmi.tag }}"
@@ -132,7 +132,7 @@ spec:
         - mountPath: /run/xtables.lock
           name: iptableslock
         - name: kubelet-config
-          mountPath: /etc/default/kubelet
+          mountPath: {{ .Values.nmi.kubeletConfig | default "/etc/default/kubelet" }}
           readOnly: true
         livenessProbe:
           httpGet:

```

aramase:
The backward compatibility tests are failing:
```
[Fail] When upgrading AAD Pod Identity [It] should be backward compatible with old and new version of MIC and NMI 
/home/vsts/work/1/s/test/e2e/framework/helm/helm_helpers.go:94
```

@maxirus Could you add a semver check similar to [this](https://github.com/Azure/aad-pod-identity/blob/master/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml#L93) for `>=1.7.0-0` while setting the  [`--kubelet-config`](https://github.com/Azure/aad-pod-identity/pull/888/files#diff-3f9c772c1f5db14acb40343123ceaa6096fd3d4d696a8fc101d7b3adc427bbebR102-R104)

maxirus:
> The backward compatibility tests are failing:
> 
> ```
> [Fail] When upgrading AAD Pod Identity [It] should be backward compatible with old and new version of MIC and NMI 
> /home/vsts/work/1/s/test/e2e/framework/helm/helm_helpers.go:94
> ```
> 
> @maxirus Could you add a semver check similar to [this](https://github.com/Azure/aad-pod-identity/blob/master/manifest_staging/charts/aad-pod-identity/templates/nmi-daemonset.yaml#L93) for `>=1.7.0-0` while setting the [`--kubelet-config`](https://github.com/Azure/aad-pod-identity/pull/888/files#diff-3f9c772c1f5db14acb40343123ceaa6096fd3d4d696a8fc101d7b3adc427bbebR102-R104)

@chewong 

Thinking about this some more; would it make more sense to just omit the flag altogether and let the K8s mount handle injecting the right hostPath config? This would allow the setting to be backwards compatible with previous versions. ie:
```yaml
#...
# Remove this:
#         {{- if and .Values.nmi.kubeletConfig (semverCompare ">= 1.7.0-0" .Values.nmi.tag) }}
#          - --kubelet-config={{ .Values.nmi.kubeletConfig }}
#          {{- end }}
#...
      volumes:
      - name: kubelet-config
        hostPath:
          path: {{ .Values.nmi.kubeletConfig }}
#...
        volumeMounts:
        - name: kubelet-config
          mountPath: /etc/default/kubelet
          readOnly: true
#...
```

Since the binary already defaults to looking for this file in `/etc/default/kubelet`, by setting `.Values.nmi.kubeletConfig` it will allow any path/file to be mounted as a static path to `/etc/default/kubelet`. This would make setting the `--kubelet-config` flag redundant.

chewong:
I think the original idea is to have identical host path and mount path but your approach makes sense. @aramase WDYT

aramase:
@maxirus That can make debugging difficult and also confuse users because of the difference in the mount location. The `--kubelet-config` flag already exists in the NMI and this is just an addition configuration option to the helm chart. As long as we bound the flag configuration with the semver check we should be good.

manifest_staging/charts/aad-pod-identity/templates/_helper.tpl
```diff
@@ -55,3 +55,14 @@ Common labels.
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 helm.sh/chart: {{ template "aad-pod-identity.chart" . }}
 {{- end -}}
+
+{{/*
+Determine KubeConfig Path
```
chewong, line 60: We can probably just wrap the kubelet-config hostPath and mountPath with the same if statement for `--kubelet-config` instead of using a template since they are not used for < v1.7.0

maxirus: I felt it was more clear and follows DRY to have a template var. Are you wanting me to remove that and put duplicate logical checks in the manifest instead?

chewong: The templating is good but let's also make sure not to mount the kubelet config path to NMI if the tag is < 1.7.0

chewong: Also, could you rename it `kubeConfig` to `kubeletConfig`

maxirus: > ...let's also make sure not to mount the kubelet config path to NMI if the tag is < 1.7.0
Not sure I follow; you we're already mounting this config path. In the template, I'm setting this path to `/etc/default/kubelet` if tag is <1.7.0.

> Also, could you rename it from `kubeConfig` to `kubeletConfig`
Sure.

chewong: The mount path is introduced in the same PR( https://github.com/Azure/aad-pod-identity/pull/842/files#diff-a6197c9515d5b54a66c30c7ffd812d27698d820afbb87671a7b0768cdf651782) so I was wondering if we can also add similar checks where we only mount kubelet config if the tag is < 1.7.0

maxirus: Ok, so you're saying that was a bug that was introduced by that PR and you want to club the fix in this PR?

chewong: Yea if that's okay with you. If not, I can open a follow-up PR to fix that

maxirus: Ok, I can add. Will also remove the template as it's redundant to check server

maxirus:
Ok, added.

IMO, these SemVer checks make the chart more complex than it needs to be. Possibly releasing a 2.0 to clean this up would be best. Additionally, keeping a static path for the KubeletConfig **inside** the container as mentioned in [this](https://github.com/Azure/aad-pod-identity/pull/888#issuecomment-738235210) comment would make debugging easier. I don't think there's an expectation in K8s that host paths must/should match container paths.

Just some things to consider.

chewong:
> Ok, added.
> 
> IMO, these SemVer checks make the chart more complex than it needs to be. Possibly releasing a 2.0 to clean this up would be best. Additionally, keeping a static path for the KubeletConfig **inside** the container as mentioned in [this](https://github.com/Azure/aad-pod-identity/pull/888#issuecomment-738235210) comment would make debugging easier. I don't think there's an expectation in K8s that host paths must/should match container paths.
> 
> Just some things to consider.

Thanks for the feedback, will definitely take into consideration when making additional changes in the future

chewong:
Going to close and open this PR to see if `license/cla` will get reset

chewong:
Looks like CLA hasn't been signed yet. What is the error you are getting?

chewong:
@maxirus any update on this PR?

PR #899, opened 2020-12-07
feat: use map for azureIdentities instead of list in helm chart

### LINKED ISSUE #885: Helm chart should consider using map rather than array for "azureIdentities"
**Describe the request**
aad-pod-identity Helm chart currently supports something like this in `values.yaml`:
```
  azureIdentities:
    - name: aso-identity
      type: 0
      # Update values for Managed Identity
      resourceID: ""
      clientID: ""
      binding:
        name: aso-identity-binding
        selector: aso_manager_binding
```

The problem is that Helm doesn't do a great job managing arrays, as you can see discussed in this issue: https://github.com/helm/helm/issues/1987, for example it seems that when using `--set azureIdentities[0].xyz` you cannot merge the contents of `values.yaml` and `--set`. If you want to say set just a few fields of an identity (maybe the `resoruceID` and `clientID`), you actually end up having to set _all_ the properties since Helm doesn't do a nice merge.

@technosophos suggested in that same thread that you could prefer a map in these instances. That in theory should allow `--set` and `values.yaml` contents to be merged like it normally is.

Would make it significantly easier to manage aad-pod-identity as a subchart, or to make use of standard `--set` and `values.yaml` merging provided by Helm.

Do note that this would be a breaking chart change I think.

aramase:
**Reason for Change**:

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

fixes #885 

**Notes for Reviewers**:

### DIFF
--- a/manifest_staging/charts/aad-pod-identity/templates/identities.yaml
+++ b/manifest_staging/charts/aad-pod-identity/templates/identities.yaml
@@ -2,27 +2,36 @@
 apiVersion: v1
 kind: List
 items:
-{{- range .Values.azureIdentities }}
+{{- range $key, $value := .Values.azureIdentities }}
   - apiVersion: "aadpodidentity.k8s.io/v1"
     kind: AzureIdentity
     metadata:
-      name: {{ .name }}
+      name: {{ $key }}
       namespace: {{ default $.Release.Namespace .namespace }}
       labels:
         {{- include "aad-pod-identity.labels" $ | nindent 8 }}
     spec:
-      type: {{ .type }}
-      resourceID: {{ required "resourceID is required!" .resourceID }}
-      clientID: {{ required "clientID is required!" .clientID }}
+      type: {{ $value.type }}
+      clientID: {{ required "clientID is required!" $value.clientID }}
+      {{- if eq (int $value.type) 0 }}
+      resourceID: {{ required "resourceID is required!" $value.resourceID }}
+      {{- else }}
+      tenantID: {{ required "tenantID is required!" $value.tenantID }}
+      clientPassword: {{ required "clientPassword is required!" $value.clientPassword }}
+      auxiliaryTenantIDs:
+        {{- range $value.auxiliaryTenantIDs }}
+        - {{ . }}
+        {{- end }}
+      {{- end }}
   - apiVersion: "aadpodidentity.k8s.io/v1"
     kind: AzureIdentityBinding
     metadata:
-      name: {{ .binding.name }}
+      name: {{ $value.binding.name }}
       namespace: {{ default $.Release.Namespace .namespace }}
       labels:
         {{- include "aad-pod-identity.labels" $ | nindent 8 }}
     spec:
-      azureIdentity: {{ .name }}
-      selector: {{ required "binding.selector is required!" .binding.selector }}
+      azureIdentity: {{ $key }}
+      selector: {{ required "binding.selector is required!" $value.binding.selector }}
 {{- end }}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


chewong (changes requested):

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -211,19 +211,30 @@ rbac:
...
-azureIdentities: []
-  # - name: "azure-identity"
+# This is a map with the AzureIdentityName being the key and the rest of the blob as value in accordance
+# to helm best practices: https://helm.sh/docs/chart_best_practices/values/#consider-how-users-will-use-your-values
+azureIdentities:
+  # "azure-identity":
   #   # if not defined, then the azure identity will be deployed in the same namespace as the chart
   #   namespace: ""
-  #   # type 0: MSI, type 1: Service Principal
+  #   # type 0: MSI, type 1: Service Principal, type 2: Service principal with certifcate
```
chewong, line 220:
```suggestion
  #   # type 0: User-assigned identity, type 1: Service Principal, type 2: Service principal with certificate
```

manifest_staging/charts/aad-pod-identity/templates/identities.yaml
```diff
@@ -2,27 +2,36 @@
...
+      clientID: {{ required "clientID is required!" $value.clientID }}
+      {{- if eq (int $value.type) 0 }}
+      resourceID: {{ required "resourceID is required!" $value.resourceID }}
+      {{- else }}
+      tenantID: {{ required "tenantID is required!" $value.tenantID }}
+      clientPassword: {{ required "clientPassword is required!" $value.clientPassword }}
+      auxiliaryTenantIDs:
+      {{- range $value.auxiliaryTenantIDs }}
+        - {{ . }}
+      {{- end }}
```
chewong, line 24: for readability:

```suggestion
        {{- range $value.auxiliaryTenantIDs }}
        - {{ . }}
        {{- end }}
```

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -211,19 +211,30 @@ rbac:
...
+  #   # Required for type 1 and 2
+  #   clientPassword: {}
+  #   # Optional for type 1 and 2 (multi-tenant)
+  #   auxiliaryTenantIDs: []
   #   binding:
   #     name: "azure-identity-binding"
   #     # The selector will also need to be included in labels for app deployment
   #     selector: "demo"
 
+
```
chewong, line 238: nit: remove extra line

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -211,19 +211,30 @@ rbac:
...
   #   type: 0
   #   # /subscriptions/subscription-id/resourcegroups/resource-group/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
+  #   # Required for type 0
   #   resourceID: ""
+  #   # Required for type 0, 1 and 2
   #   clientID: ""
+  #   # Required for type 1 and 2
+  #   tenantID: ""
+  #   # Required for type 1 and 2
+  #   clientPassword: {}
```
chewong, line 230:
```suggestion
  #   clientPassword:
  #     name: ""
  #     namespace: "" 
```

chewong (approved):
LGTM with a few nits

manifest_staging/charts/aad-pod-identity/values.yaml
aramase: `clientPassword` is not a map. It is [types.NamespacedName](https://pkg.go.dev/k8s.io/apimachinery/pkg/types#NamespacedName). I've updated the PR with the format.

PR #937, opened 2021-01-12
fix: check VM/VMSS state before PATCH calls

### LINKED ISSUE #926: Check VMSS state before PATCH calls
**Describe the request**

**Additional context**

chewong:

**Reason for Change**:

Fixes #926.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/cloudprovider/vm.go
+++ b/pkg/cloudprovider/vm.go
@@ -90,7 +90,7 @@ func (c *VMClient) Get(rgName string, nodeName string) (compute.VirtualMachine, 
 		resp := vm.Response.Response
 		// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
 		c.retryAfterReader = time.Now().Add(getRetryAfter(resp))
-		return vm, fmt.Errorf("failed to get vm %s in resource group %s, error: %+v", nodeName, rgName, err)
+		return compute.VirtualMachine{}, fmt.Errorf("failed to get vm %s in resource group %s, error: %+v", nodeName, rgName, err)
 	}
 	stats.Increment(stats.TotalGetCalls, 1)
 	stats.AggregateConcurrent(stats.CloudGet, begin, time.Now())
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/vm.go
```diff
@@ -99,6 +99,10 @@ func (c *VMClient) Get(rgName string, nodeName string) (compute.VirtualMachine,
 
 // UpdateIdentities updates the user assigned identities for the provided node
 func (c *VMClient) UpdateIdentities(rg, nodeName string, vm compute.VirtualMachine) error {
+	if *vm.ProvisioningState != string(compute.ProvisioningStateSucceeded) {
```
chewong, line 102: Let me know if there are other states that we want to check

aramase: I think just waiting for `ProvisioningStateSucceeded` might cause delay in the identity assignment operations. We should check if the `ProvisioningState` is `Deleting` and skip operations only for those.

pkg/cloudprovider/vm.go
```diff
@@ -99,6 +99,10 @@ func (c *VMClient) Get(rgName string, nodeName string) (compute.VirtualMachine,
 
 // UpdateIdentities updates the user assigned identities for the provided node
 func (c *VMClient) UpdateIdentities(rg, nodeName string, vm compute.VirtualMachine) error {
+	if *vm.ProvisioningState == string(compute.ProvisioningState1Deleting) {
```
aramase, line 102: Why is this `ProvisioningState1Deleting` instead of `ProvisioningStateDeleting`?

chewong: fixed

pkg/cloudprovider/vmss.go
```diff
@@ -136,7 +140,7 @@ func (c *VMSSClient) Get(rgName string, vmssName string) (ret compute.VirtualMac
 		resp := vmss.Response.Response
 		// Update RetryAfterReader so that no more requests would be sent until RetryAfter expires.
 		c.retryAfterReader = time.Now().Add(getRetryAfter(resp))
-		return vmss, fmt.Errorf("failed to get vmss %s in resource group %s, error: %+v", vmssName, rgName, err)
+		return compute.VirtualMachineScaleSet{}, fmt.Errorf("failed to get vmss %s in resource group %s, error: %+v", vmssName, rgName, err)
```
aramase, line 143: nit: We should also make this change for `vm.go`

aramase (approved):

PR #981, opened 2021-03-01
feat: add cluster identity to immutable list

### LINKED ISSUE #979: Skip identity deletion if identity is being used by MIC
**Describe the request**
Explore adding a check in MIC to skip un-assigning identity with the same `clientID` as the identity it is using for all Azure operations.

**Additional context**

aramase:

**Reason for Change**:

If the cluster identity is used with the application pod, then when all pods using the identity are deleted, the cluster identity will get cleaned up by MIC. This will cause subsequent cloud provider operations to fail because the cluster identity no longer exists. The user would need to manually go and fix it by assigning the identity back. Instead this PR adds the cluster identity to the immutable identity list. The cluster identity will never be deleted by MIC.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

fixes #979 

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/node_test.go
+++ b/test/e2e/node_test.go
@@ -75,9 +75,9 @@ var _ = Describe("When managing identities from the underlying nodes", func() {
 	It("should not delete a user-assigned identity that is being used by a different pod", func() {
 		// This test is specifically testing VMSS behavior
 		// As such we'll look through the cluster to see if there are nodes assigned
-		// to a VMSS, and if any of thoe VMSS's have more than one node.
+		// to a VMSS, and if any of those VMSS's have more than one node.
 		//
-		// We cannot do the test if there is not at least1 VMSS with at least 2 nodes
+		// We cannot do the test if there is not at least 1 VMSS with at least 2 nodes
 		// attach to it.
 		var vmssNodes []corev1.Node
 		if vmssNodes = getVMSSNodes(nodes); len(vmssNodes) < 2 {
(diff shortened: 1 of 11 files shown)


chewong (commented):
Could you rebase it so this PR uses the updated CI config?

chewong (commented):
Is it possible to add a simple e2e test? We should be able to fetch the client ID of the cluster identity using [`GetIdentityClientID`](https://github.com/Azure/aad-pod-identity/blob/master/test/e2e/framework/azure/azure_client.go#L78) and create an azure-cli pod perform a simple `az login -i -u <resource id>`.

chewong:
Also, feel free to ignore the nightly CI signal. I misconfigured it and it's now showing up in this PR

aramase:
> Is it possible to add a simple e2e test? We should be able to fetch the client ID of the cluster identity using GetIdentityClientID and create an azure-cli pod perform a simple az login -i -u <resource id>, and then making sure the identity is still assigned to the node after the pod is deleted.

Yeah, I'm exploring that option. I think the general idea will be to run the test if there is an agentpool identity (this name is deterministic in our CI) assigned. If not, skip the test.

test/e2e/node_test.go
```diff
@@ -420,8 +484,20 @@ func getVMSSNodes(nodes *corev1.NodeList) []corev1.Node {
...
+	resourceID := fmt.Sprintf(azure.ResourceIDTemplate, config.SubscriptionID, resourceGroup, identityToCheck)
 	_, ok := identities[strings.ToLower(resourceID)]
 	return ok
 }
+
+// getClusterIdentityName returns the name of cluster identity that
+// contains the agentpool suffix
+func getClusterIdentityName(identities map[string]bool) string {
+	for resourceID := range identities {
+		if strings.Contains(resourceID, "-agentpool") {
```
chewong, line 497:
```suggestion
		if strings.HasSuffix(resourceID, "-agentpool") {
```

chewong (approved):

PR #1035, opened 2021-04-08
chore: promote crd to apiextensions.k8s.io/v1 and remove role assignments after e2e test

### LINKED ISSUE #661: Upgrade CRDs to apiextensions.k8s.io/v1
CRDs have been promoted to GA since Kubernetes 1.16:

https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20180415-crds-to-ga.md

`aad-pod-identity` CRDs currently use apiextensions.k8s.io/v1beta1, which will be removed in the future: kubernetes/kubernetes#82022.

**Note:**

- The removal of `apiextensions.k8s.io/v1beta1` is planned for 1.22. So `aad-pod-identity` will continue to use `apiextensions.k8s.io/v1beta1` until the 1.21 release in order to support clusters < 1.16. Before the 1.22 release, we can make the switch to `apiextensions.k8s.io/v1` and set the default kube version of helm chart to 1.16+

**Alternatives:**

helm3 doesn't support templating in the crds dir. If this is possible, then we can determine the k8s version and use the appropriate `apiVersion`.

### LINKED ISSUE #994: ci: remove role assignments from aad-pod-identity-artifacts resource group
dangling role assignments are represented as `Identity not found` entry in the Azure portal
image
We will need to remove them after finish running the e2e test.

chewong:

**Reason for Change**:

fixes #661
fixes #994 

- use controller-gen to generate YAML files for our CRDs
- existing users with 1.16+ clusters should not be affected by this PR since kube-apiserver already converts CRDs with apiVersion of v1beta1 to v1 automatically
- helm 3 does not re-configure the CRDs during an upgrade (meaning new fields like openAPIV3Schema won't be added), whereas `kubectl apply -f` does re-configure the CRDs to include new fields added in this PR
- existing CRDs data should be persisted unless the user explicitly runs `kubectl delete crds`

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:


test/e2e/backward_compat_test.go
```diff
@@ -1,138 +0,0 @@
-// +build e2e
```
chewong, line 1: Removing this test due to v1 CRDs validation, which does not allow unknown fields to exist in the spec.

chewong:
@aramase this PR is ready for review

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -232,7 +232,7 @@ nmi:
 
   # enable running aad-pod-identity on clusters with kubenet
   # default is false
-  allowNetworkPluginKubenet: false
+  allowNetworkPluginKubenet: true
```
aramase, line 235: Was this an change intended?

pkg/apis/aadpodidentity/v1/types.go
```diff
@@ -185,48 +187,51 @@ const (
...
 // AzureAssignedIdentitySpec contains the relationship
 // between an AzureIdentity and an AzureIdentityBinding.
 type AzureAssignedIdentitySpec struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	AzureIdentityRef  *AzureIdentity        `json:"azureIdentityRef"`
-	AzureBindingRef   *AzureIdentityBinding `json:"azureBindingRef"`
-	Pod               string                `json:"pod"`
-	PodNamespace      string                `json:"podNamespace"`
-	NodeName          string                `json:"nodename"`
+	// +kubebuilder:validation:XEmbeddedResource
```
aramase, line 206: Could we also add a comment here on why we are adding this?

test/e2e/backward_compat_test.go
aramase: sounds good!

manifest_staging/charts/aad-pod-identity/values.yaml
chewong: Good catch. I turned this on for testing but forgot to revert the change

aramase (approved):

PR #1036, opened 2021-04-08
helm: ability to add AzureIdentities with the same name across different namespaces

### LINKED ISSUE #1020: Can't register identity in multiple namespaces through helm values for v3
**Describe the bug**

Because of the nature itself of maps a key cannot be used twice and hence it is not possible to create the same identity in two different namespaces by using the azureIdentities attribute of the helm chart.

```
+ azureIdentities:
+  "azure-identity": # If we attempt to provide multiple with different namespace value only last will get created
+    # if not defined, then the azure identity will be deployed in the same namespace as the chart
+    namespace: "namespace1"
+    # type 0: MSI, type 1: Service Principal
+    type: 0
+    # /subscriptions/subscription-id/resourcegroups/resource-group/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
+    resourceID: "resource-id"
+    clientID: "client-id"
+    binding:
+      name: "azure-identity-binding"
+      # The selector will also need to be included in labels for app deployment
+      selector: "demo"
```

**Expected behavior**

One should be able to create an identity for each namespace when working in namespaced identities mode.

One could get back the ability to do so by:

```
 azureIdentities:
  "azure-identity": # If we attempt to provide multiple with different namespace value only last will get created
    # if not defined, then the azure identity will be deployed in the same namespace as the chart
-   namespace:
+   namespaces: 
+    - namespace1
+    - namespace2
    # type 0: MSI, type 1: Service Principal
    type: 0
    # /subscriptions/subscription-id/resourcegroups/resource-group/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
    resourceID: "resource-id"
    clientID: "client-id"
    binding:
      name: "azure-identity-binding"
      # The selector will also need to be included in labels for app deployment
      selector: "demo"
```

or:

```
 azureIdentities:
  "azure-identity-namespace1": 
    # if not defined, then the azure identity will be deployed in the same namespace as the chart
    namespace: "namespace1"
+   name: "azure-identity" # We can optionally provide the name of the identity as a parameter. Defalt will take the key as name
    # type 0: MSI, type 1: Service Principal
    type: 0
    # /subscriptions/subscription-id/resourcegroups/resource-group/providers/Microsoft.ManagedIdentity/userAssignedIdentities/identity-name
    resourceID: "resource-id"
    clientID: "client-id"
    binding:
      name: "azure-identity-binding"
      # The selector will also need to be included in labels for app deployment
      selector: "demo"
```

**AAD Pod Identity version**
1.7.4
**Kubernetes version**
1.20.2
**Additional context**

chewong:

**Reason for Change**:

Fixes #1020 

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/manifest_staging/charts/aad-pod-identity/templates/identities.yaml
+++ b/manifest_staging/charts/aad-pod-identity/templates/identities.yaml
@@ -6,8 +6,8 @@ items:
   - apiVersion: "aadpodidentity.k8s.io/v1"
     kind: AzureIdentity
     metadata:
-      name: {{ $key }}
-      namespace: {{ default $.Release.Namespace .namespace }}
+      name: {{ default $key $value.name }}
+      namespace: {{ default $.Release.Namespace $value.namespace }}
       labels:
         {{- include "aad-pod-identity.labels" $ | nindent 8 }}
     spec:
@@ -27,11 +27,11 @@ items:
     kind: AzureIdentityBinding
     metadata:
       name: {{ $value.binding.name }}
-      namespace: {{ default $.Release.Namespace .namespace }}
+      namespace: {{ default $.Release.Namespace $value.namespace }}
       labels:
         {{- include "aad-pod-identity.labels" $ | nindent 8 }}
     spec:
-      azureIdentity: {{ $key }}
+      name: {{ default $key $value.name  }}
       selector: {{ required "binding.selector is required!" $value.binding.selector }}
 {{- end }}
 {{- end -}}
--- a/manifest_staging/charts/aad-pod-identity/README.md
+++ b/manifest_staging/charts/aad-pod-identity/README.md
@@ -182,6 +182,8 @@ The following is a basic example of the required change in the user-supplied val
 -      selector: "demo"
 +azureIdentities:
 +  "azure-identity":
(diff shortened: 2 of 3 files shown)


manifest_staging/charts/aad-pod-identity/templates/identities.yaml
```diff
@@ -6,7 +6,7 @@ items:
   - apiVersion: "aadpodidentity.k8s.io/v1"
     kind: AzureIdentity
     metadata:
-      name: {{ $key }}
+      name: {{ default $key $value.name }}
       namespace: {{ default $.Release.Namespace .namespace }}
```
aramase, line 10: Shouldn't this also be `{{ default $.Release.Namespace $value.namespace }}`

chewong: Looks like both work but I can apply your change to make it clearer

aramase (approved):

PR #1043, opened 2021-04-13
Create optional user roles for AAD Pod Identity

### LINKED ISSUE #1018: Helm chart: Provide user facing cluster roles
I have seen some Helm charts for popular applications provide user facing cluster roles which allow various operations on the custom CRD resources which usually aggregate into the built-in Kubernetes admin, edit and view roles.

This would help in situations where you have service accounts for deployments and you want to either bind specific cluster roles or use the built-in Kubernetes namespace admin/edit roles to help secure the deployment accounts. 

Right now we solve this on our clusters by creating a custom role with access to the relevant CRDs for AAD Pod Identity as well as some other apps which don't provide user facing roles and bind them to the service accounts we create.

I think we would just need 2 roles, one which would aggregate to the admin and edit roles and allows full access to the CRDs, and a view role which aggregates to the view role. Currently our custom role allows access to the following resources:

- AzureIdentity
- AzureIdentityBinding

adhodgson1:
**Reason for Change**:
This PR optionally creates cluster roles for view and edit resources used by the AAD Pod Identity components. They are typically used to provide CI processes access to create the necessary resources to allow applications to use AAD Pod Identity.

**Requirements**

- [X] squashed commits
- [X] included documentation
- [X] ran `make precommit`

**Issue Fixed**:
Fixes #1018


manifest_staging/charts/aad-pod-identity/README.md
```diff
@@ -291,6 +291,7 @@ The following tables list the configurable parameters of the aad-pod-identity ch
...
 | `rbac.pspEnabled`                         | If `true`, create and use a restricted pod security policy for AAD Pod Identity pod(s)                                                                                                                                                                                                                                        | `false`                                                        |
 | `rbac.allowAccessToSecrets`               | NMI requires permissions to get secrets when service principal (type: 1) is used in AzureIdentity. If using only MSI (type: 0) in AzureIdentity, secret get permission can be disabled by setting this to false.                                                                                                              | `true`                                                         |
+| `rbac.createUserFacingClusterRoles` | If set to true, then view and edit cluster roles will be created with annotations that agrigate to the admin, edit and view built-in Kubernetes cluster roles. These roles will be able to create the necessary resources to allow pod identity binding on pods. | `true` |
```
aramase, line 294:
```suggestion
| `rbac.createUserFacingClusterRoles` | If set to true, then view and edit cluster roles will be created with annotations for the admin, edit and view built-in Kubernetes cluster roles. These roles will be able to create the necessary resources to allow pod identity binding on pods. | `false` |
```

aramase: The default value should be `false`

manifest_staging/charts/aad-pod-identity/templates/user-facing-clusterroles.yaml
```diff
@@ -0,0 +1,28 @@
...
+metadata:
+  name: {{ template "aad-pod-identity.fullname" . }}-view
+  labels:
+    {{- include "aad-pod-identity.labels" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+  - apiGroups: ["aadpodidentity.k8s.io"]
+    resources: ["azureidentitybindings", "azureidentities", "azureassignedidentities", "azurepodidentityexceptions"]
```
aramase, line 13: If pod identity is run in [managed mode](https://azure.github.io/aad-pod-identity/docs/configure/pod_identity_in_managed_mode/), then we don't permissions on `azureassignedidentities`. We should have a conditional check here based on the mode

https://github.com/Azure/aad-pod-identity/blob/cee87b5f1ee25558498632bca1f6da27e4eb3d30/charts/aad-pod-identity/values.yaml#L40-L42

manifest_staging/charts/aad-pod-identity/README.md
adhodgson1: Thanks it was just set to true in the docs, fixed.

manifest_staging/charts/aad-pod-identity/templates/user-facing-clusterroles.yaml
adhodgson1: I have put the conditional check on the view role so it only applies the read on azureassignedidentities if the operational mode is standard. I have removed azureassignedidentities from the edit role as I don't believe users will ever need to update these, is that a correct assumption?

aramase: That's correct. `AzureAssignedIdentity` is created by MIC and users shouldn't be able to modify it.

aramase (approved):

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -244,6 +244,10 @@ rbac:
   # If using only MSI (type: 0) in AzureIdentity, secret get permission can be disabled by setting this to false.
   allowAccessToSecrets: true
   pspEnabled: false
+  # If set to true, then view and edit cluster roles will be created with annotations 
+  # that agrigate to the admin, edit and view built-in cluster roles. These roles will
```
aramase, line 248: small typo, but we can fix this in another PR

chewong (approved):

aramase:
@adhodgson1 Thank you for the PR!

Issue #830, opened 2020-10-13
Show more information in kubectl get AzureIdentityBinding


tomkerkhove:
**Describe the request**
Show more information in kubectl get AzureIdentityBinding to help operators get an understanding of what is provisioned when using `-o wide`

**Explain why AAD Pod Identity needs it**
Make it more operable and easier to use

**Describe the solution you'd like**
Information concerning `azureIdentity` & `selector`

tomkerkhove:
keep-alive

chewong:
Fixed by #1093

Issue #829, opened 2020-10-13
Show more information in kubectl get AzureIdentity


tomkerkhove:
**Describe the request**
Show more information in kubectl get AzureIdentity to help operators get an understanding of what is provisioned when using `-o wide`

**Explain why AAD Pod Identity needs it**
Make it more operable and easier to use

**Describe the solution you'd like**
Information concerning `type`, `resourceID` & `clientID`

chewong:
Fixed by #1093

tomkerkhove:
Thanks!

PR #1093, opened 2021-06-14
Add additional columns to kubectl output

JasonWhall:
**Reason for Change**:

Adds additional columns to kubectl output for AAD Pod Identity Custom Resources.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- #829 
- #830 

- [ ] yes
- [x] no

### DIFF
--- a/pkg/apis/aadpodidentity/v1/types.go
+++ b/pkg/apis/aadpodidentity/v1/types.go
@@ -31,6 +31,9 @@ const (
 
 // AzureIdentity is the specification of the identity data structure.
 //+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//+kubebuilder:printcolumn:name="Type",type="string",JSONPath=".spec.type",description="",priority=0
+//+kubebuilder:printcolumn:name="ClientID",type="string",JSONPath=".spec.clientID",description="",priority=0
+//+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC."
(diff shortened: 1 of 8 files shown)


aramase (commented):
@JasonWhall Thank you for adding this. Could you also add this to the CRD definitions in [manifest_staging/deploy/infra](https://github.com/Azure/aad-pod-identity/tree/master/manifest_staging/deploy/infra) folder?

chewong (commented):
We use `kubebuilder` to generate CRDs Could you:
- check out https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns on how to add additional columns via kubebuilder markers on the Go type
- `make generate-crds` to generate CRDs after making the changes

JasonWhall:
@chewong @aramase I've updated the kubebuilder flags and hopefully correctly generated the crds and put the result in the right place! There's a bit of noise in the files, assuming this is ok.

Thanks, TIL ❤️

pkg/apis/aadpodidentity/v1/types.go
```diff
@@ -31,6 +31,9 @@ const (
 
 // AzureIdentity is the specification of the identity data structure.
 //+k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//+kubebuilder:printcolumn:name="Type",type="string",JSONPath=".spec.type",description="",priority=1
```
chewong, line 34: I am thinking maybe we can assign the priority to 0 so we don't have to append `-owide` during `kubectl get` to get the additional information. What do you think?

JasonWhall: Yep, makes sense to me as they should be present on all resources of this type 👍 I've updated the PR to reflect this.

chewong (approved):
This is great! Thanks for working on this

aramase (approved):
Thank you for PR!

lgtm

PR #1098, opened 2021-06-22
fix: mount kubelet config to `/var/lib/kubelet` for non-rbac deployment

bcho:
**Reason for Change**:

AKS is rolling out the TLS bootstrapping change. Before this change, AKS cluster agent node uses kubeconfig as follow:

```yaml
apiVersion: v1
kind: Config
clusters:
- name: localcluster
  cluster:
    certificate-authority: /etc/kubernetes/certs/ca.crt
    server: https://<api-server-address>:443
users:
- name: client
  user:
    client-certificate: /etc/kubernetes/certs/client.crt
    client-key: /etc/kubernetes/certs/client.key
contexts:
- context:
    cluster: localcluster
    user: client
  name: localclustercontext
current-context: localclustercontext
```

After enabling TLS bootstrapping, the agent node will get the kubeconfig as follow:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority: /etc/kubernetes/certs/ca.crt
    server: https://<api-server-address>:443
  name: default-cluster
contexts:
- context:
    cluster: default-cluster
    namespace: default
    user: default-auth
  name: default-context
current-context: default-context
kind: Config
preferences: {}
users:
- name: default-auth
  user:
    client-certificate: /var/lib/kubelet/pki/kubelet-client-current.pem
    client-key: /var/lib/kubelet/pki/kubelet-client-current.pem
```

As we can see, in the new mode, kubelet's kubeconfig references the client certificates in `/var/lib/kubelet/pki` instead of `/etc/kubernetes/certs`. 

Since we are mounting the kubeconfig from host for those RBAC-disabled cluster, I think it's better to maintain the mount path structure for the kubeconfig (i.e. mounting `/var/lib/kubelet` from host to `/var/lib/kubelet` inside the pod) - with this, we can avoiding breaking MIC for certificates change.

I had verified this change in two AKS RBAC-disabled clusters, using with/out TLS bootstrapping setup.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,9 +17,12 @@ RUN export GOOS=$TARGETOS && \
     export GOARM=$(echo ${TARGETPLATFORM} | cut -d / -f3 | tr -d 'v') && \
     make build
 
-FROM k8s.gcr.io/build-image/debian-iptables:buster-v1.6.0 AS nmi
-# upgrading liblz4-1 due to CVE-2021-3520
-RUN clean-install ca-certificates liblz4-1
+FROM k8s.gcr.io/build-image/debian-iptables:buster-v1.6.3 AS nmi
+# upgrading libgcrypt20 due to CVE-2021-33560
+# upgrading libgnutls30 due to CVE-2021-20231, CVE-2021-20232, CVE-2020-24659
+# upgrading libhogweed4 due to CVE-2021-20305, CVE-2021-3580
+# upgrading libnettle6 due to CVE-2021-20305, CVE-2021-3580
+RUN clean-install ca-certificates libgcrypt20 libgnutls30 libhogweed4 libnettle6
 COPY --from=builder /go/src/github.com/Azure/aad-pod-identity/bin/aad-pod-identity/nmi /bin/
 RUN useradd -u 10001 nonroot
 USER nonroot
--- a/manifest_staging/deploy/infra/deployment.yaml
+++ b/manifest_staging/deploy/infra/deployment.yaml
@@ -496,7 +496,7 @@ spec:
       - name: mic
         image: "mcr.microsoft.com/oss/azure/aad-pod-identity/mic:v1.8.0"
         args:
-          - "--kubeconfig=/etc/kubernetes/kubeconfig/kubeconfig"
+          - "--kubeconfig=/var/lib/kubelet/kubeconfig"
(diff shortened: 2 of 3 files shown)


aramase (commented):
@bcho All the changes are to be made in [manifest_staging/deploy](https://github.com/Azure/aad-pod-identity/tree/master/manifest_staging/deploy) folder and will be promoted to `deploy` folder at the time of release.

deploy/infra/deployment.yaml
```diff
@@ -490,8 +490,8 @@ spec:
             cpu: 100m
             memory: 256Mi
         volumeMounts:
-          - name: kubeconfig
-            mountPath: /etc/kubernetes/kubeconfig
+          - name: kubelet-config
+            mountPath: /var/lib/kubelet
```
aramase, line 494: Why is this change required? The volume mount is only for within the pod. The actual volume that's still mounted is `/var/lib/kubelet` hostPath.

bcho: I put the background in the PR description. The main reason is with TLS bootstrapping enabled, this kubeconfig will reference the client certificates in `/var/lib/kubelet/pki`. In this case, we should mount the whole `/var/lib/kubelet` into the pod.

bcho:
Hi @aramase I updated the PR to modify the staging folder only, do we need to run `promote-staging-manifest` to promote the changes?

aramase:
> Hi @aramase I updated the PR to modify the staging folder only, do we need to run promote-staging-manifest to promote the changes?

1. Could you also update https://github.com/Azure/aad-pod-identity/blob/master/manifest_staging/deploy/infra/noazurejson/deployment.yaml?
2. `promote-staging-manifest` will be run at the time of the release. So it's not required now. We'll include this yaml update in the `v1.8.1` release.

bcho:
@aramase updated, PTAL

aramase (approved):
> I had verified this change in two AKS RBAC-disabled clusters, using with/out TLS bootstrapping setup.

@bcho Thanks for confirming the change is backward compatible.

lgtm

chewong (approved):

PR #1099, opened 2021-06-22
chore: enable scale features by default

aramase:

**Reason for Change**:

- Enables scale features by default for NMI to reduce memory consumption and # of API calls by default
- Deprecates unused scale features flag in MIC. This flag will be removed in the release after `v1.8.1`

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/manifest_staging/charts/aad-pod-identity/values.yaml
+++ b/manifest_staging/charts/aad-pod-identity/values.yaml
@@ -218,8 +218,8 @@ nmi:
   findIdentityRetryIntervalInSeconds: ""
 
   # Enable scale features - https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.featureflags.md#enable-scale-features-flag
-  # Accepted values are true/false. Default is false.
-  enableScaleFeatures: ""
+  # Accepted values are true/false. Default is true for v1.8.1+.
+  enableScaleFeatures: true
 
   # default value is 9090
   # prometheus port for metrics
--- a/manifest_staging/deploy/infra/deployment-rbac.yaml
+++ b/manifest_staging/deploy/infra/deployment-rbac.yaml
@@ -479,6 +479,7 @@ spec:
         args:
           - "--node=$(NODE_NAME)"
           - "--http-probe-port=8085"
+          - "--enableScaleFeatures=true"
         env:
           - name: HOST_IP
             valueFrom:
--- a/cmd/mic/main.go
+++ b/cmd/mic/main.go
@@ -130,7 +130,7 @@ func main() {
 	}
 
 	if enableScaleFeatures {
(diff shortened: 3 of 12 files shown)


chewong (commented):
Can we have `enableScaleFeatures = pflag.Bool("enableScaleFeatures", true, "Enable/Disable features for scale clusters")` in `cmd/nmi/main.go` so users who upgrade with `kubectl edit daemonset` from v1.8.0 to v1.8.1 can have this feature enabled by default?

aramase:
> Can we have enableScaleFeatures = pflag.Bool("enableScaleFeatures", true, "Enable/Disable features for scale clusters") in cmd/nmi/main.go so users who upgrade with kubectl edit daemonset from v1.8.0 to v1.8.1 can have this feature enabled by default?

I do have that change and I see it locally. Let me force push it.

manifest_staging/deploy/infra/deployment-rbac.yaml
```diff
@@ -479,6 +479,7 @@ spec:
         args:
           - "--node=$(NODE_NAME)"
           - "--http-probe-port=8085"
+          - "--enableScaleFeatures=true"
```
chewong, line 482: Can we remove this since it's true by default?

manifest_staging/charts/aad-pod-identity/values.yaml
```diff
@@ -218,8 +218,8 @@ nmi:
   findIdentityRetryIntervalInSeconds: ""
 
   # Enable scale features - https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.featureflags.md#enable-scale-features-flag
-  # Accepted values are true/false. Default is false.
-  enableScaleFeatures: ""
+  # Accepted values are true/false. Default is true for v1.8.1+.
+  enableScaleFeatures: true
```
chewong, line 222: ditto

manifest_staging/deploy/infra/deployment-rbac.yaml
aramase: I think we should keep it until the flag is supported so user's can just disable when they want to without looking at the docs.

manifest_staging/charts/aad-pod-identity/values.yaml
aramase: The default is true but user's can still disable if they want to. We can remove this configuration later if we decide to not support disabling it. WDYT?

chewong: Sounds good

chewong (approved):

PR #1114, opened 2021-07-21
feat: option to set Retry-After in NMI responses

aramase:
Adds a new feature flag to enable setting Retry-After header in the error response from NMI. The error is only when the identity is still being assigned by NMI or no valid AzureAssignedIdentity is found yet. This enables SDK's to retry based on the http status code 503 and the retry after header.

Note: When enabling this feature, the default retries in NMI should be explicitly disabled to rather rely on the SDK for retries.

`--retry-attempts-for-created=1,` `--retry-attempts-for-assigned=1` and `--find-identity-retry-interval=1`. This will force NMI to return immediately with the Retry-After header 20s.

**Reason for Change**:

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:


aramase:
Unit test failure should be fixed with https://github.com/Azure/aad-pod-identity/pull/1116

pkg/nmi/server/server.go
```diff
@@ -373,7 +376,14 @@ func (s *Server) msiHandler(w http.ResponseWriter, r *http.Request) (ns string)
 	podID, err := s.TokenClient.GetIdentities(r.Context(), podns, podname, tokenRequest.ClientID, tokenRequest.ResourceID)
 	if err != nil {
 		klog.Errorf("failed to get matching identities for pod: %s/%s, error: %+v", podns, podname, err)
-		http.Error(w, err.Error(), http.StatusNotFound)
+		httpErrorCode := http.StatusFound
```
chewong, line 379: Is this supposed to be `http.StatusNotFound`

aramase: That is the default value today. We're keeping the default behavior as is. If the feature is enabled, then we'll set it to `http.StatusServiceUnavailable` and add the `Retry-After` header.

aramase: Updated!

chewong (approved):

PR #1150, opened 2021-08-27
docs: specify the latest version number containing the breaking changes

### LINKED ISSUE #1130: Please specify the latest version number containing breaking changes
**Have you**

- [X] Read [Troubleshooting Guide](../../docs/readmes/README.troubleshooting.md)
- [X] Read [Known Issues](../../docs/readmes/README.known-issues.md)
- [X] Searched on [GitHub issues](https://github.com/Azure/aad-pod-identity/issues)

As best I can tell, the [most recent breaking change was in 1.7.5](https://azure.github.io/aad-pod-identity/changelog/#v182), so I don't understand why release notes since then each say "v1.8.x contains breaking changes". That's true for anyone upgrading from before 1.7.5, but by that logic, you'll never ship another version that doesn't include breaking changes, which would be crying wolf.

Please specify the latest version number with a breaking change, rather than relying on the wildcard, to allow people to do the math when upgrading. Thanks for your time.

chewong:

**Reason for Change**:

Fixes #1130.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/website/content/en/changelog/_index.md
+++ b/website/content/en/changelog/_index.md
@@ -7,31 +7,6 @@ menu:
     weight: 10
 ---
 
-
-### Breaking Change
-
-- If upgrading from versions 1.5.x to 1.7.x of pod-identity, please carefully review this [doc](https://azure.github.io/aad-pod-identity/docs/#v16x-breaking-change) before upgrade.
-- Pod Identity is disabled by default for Clusters with Kubenet. Please review this [doc](https://azure.github.io/aad-pod-identity/docs/configure/aad_pod_identity_on_kubenet/) before upgrade.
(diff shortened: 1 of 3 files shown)


aramase (commented):
The updates to changelog looks good but do we want to remove it from the main docs? If we remove the breaking changes summarized between minor versions then the user would need to review the changelog for every single release before upgrade. Should we also keep the main breaking changes section and add the version introduced for each change?

chewong:
Added back the breaking change section. PTAL.

aramase (approved):

PR #1206, opened 2021-12-10
feat: return http 503 when IMDS healthcheck fails 

aramase:
**Reason for Change**:

- Parse the error from adal token request to distinguish token request error from IMDS health check failures.
  - If IMDS health check fails, return a 503
  - For token request failures, keep the same behavior and return 403
- Update the token request method in all the packages as the current ones we're using are deprecated.

**Requirements**

- [x] squashed commits
- [ ] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -3,14 +3,12 @@ package auth
 import (
 	"context"
 	"crypto/rsa"
-	"fmt"
 	"time"
 
 	"github.com/Azure/aad-pod-identity/pkg/metrics"
 	"github.com/Azure/aad-pod-identity/version"
 
 	"github.com/Azure/go-autorest/autorest/adal"
-
 	"golang.org/x/crypto/pkcs12"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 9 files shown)


pkg/nmi/server/server.go
```diff
@@ -245,7 +245,14 @@ func (s *Server) hostHandler(w http.ResponseWriter, r *http.Request) (ns string)
 	tokens, err := s.TokenClient.GetTokens(r.Context(), tokenRequest.ClientID, tokenRequest.Resource, *podID)
 	if err != nil {
 		klog.Errorf("failed to get service principal token for pod:%s/%s, error: %+v", podns, podname, err)
-		http.Error(w, err.Error(), http.StatusForbidden)
+		httpErrorCode := http.StatusForbidden
+		if !auth.IsTokenRefreshError(err) {
```
chewong, line 249: From what I understand, if it's not a token refresh error, it's a health check error. Can we rename the function name to `IsHealthCheckError` to make it more explicit?

pkg/auth/auth.go
```diff
@@ -38,19 +36,14 @@ func GetServicePrincipalTokenFromMSI(resource string) (_ *adal.Token, err error)
...
-	msiEndpoint, err := adal.GetMSIVMEndpoint()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get the MSI endpoint, error: %+v", err)
-	}
-	// Set up the configuration of the service principal
-	spt, err := adal.NewServicePrincipalTokenFromMSI(msiEndpoint, resource)
+	spt, err := adal.NewServicePrincipalTokenFromManagedIdentity(resource, nil)
 	if err != nil {
-		return nil, fmt.Errorf("failed to acquire a token for MSI, error: %+v", err)
+		return nil, err
```
chewong, line 41: Is there a reason why we remove the description of the error? I found them pretty helpful when debugging and determining where the error is coming from.

pkg/nmi/server/server.go
aramase: With the current sdk thats a valid assumption because every sdk call is health check + token request. I can add `IsHealthCheckError` and internally call `IsTokenRefreshError` because token refresh error is the one that we can actually deduce.

pkg/auth/auth.go
aramase: From initial glance, it looks like we're wrapping the error at every stage so decided to remove it from here. I can add it back for now.

pkg/nmi/server/server.go
chewong: Sounds good

pkg/auth/auth.go
chewong: Didn't realize that. Feel free to remove those that are duplicated. Thanks!

chewong (approved):
Feel free to merge after rebasing.

PR #1235, opened 2022-01-16
feat: Add AllowedNamespaceToInstanceMetadata flag to NMI

adriananeci:
**Reason for Change**:
Add new flag for NMI, `AllowedNamespaceToInstanceMetadata`, in order to allow pods from specific namespace to access the instance metadata when `blockInstanceMetadata` flag is set.

A use case will be to allow Azure CSI drivers to access instance metadata when running in overlay network.

**Requirements**

- [x] squashed commits
- [x] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/cmd/nmi/main.go
+++ b/cmd/nmi/main.go
@@ -45,6 +45,7 @@ var (
 	enableProfile                      = pflag.Bool("enableProfile", false, "Enable/Disable pprof profiling")
 	enableScaleFeatures                = pflag.Bool("enableScaleFeatures", true, "Enable/Disable features for scale clusters")
 	blockInstanceMetadata              = pflag.Bool("block-instance-metadata", false, "Block instance metadata endpoints")
+	allowedNamespaceToInstanceMetadata = pflag.String("allowed-namespace-to-instance-metadata", "kube-system", "Allowed namespace to instance metadata endpoints if `block-instance-metadata` flag is set")
 	metadataHeaderRequired             = pflag.Bool("metadata-header-required", true, "Metadata header required for querying Azure Instance Metadata service")
 	prometheusPort                     = pflag.String("prometheus-port", "9090", "Prometheus port for metrics")
 	operationMode                      = pflag.String("operation-mode", "standard", "NMI operation mode")
@@ -113,7 +114,7 @@ func main() {
 	*forceNamespaced = *forceNamespaced || "true" == os.Getenv("FORCENAMESPACED")
 	klog.Infof("running NMI in namespaced mode: %v", *forceNamespaced)
 
-	s := server.NewServer(*micNamespace, *blockInstanceMetadata, *metadataHeaderRequired, *setRetryAfterHeader)
+	s := server.NewServer(*micNamespace, *allowedNamespaceToInstanceMetadata, *blockInstanceMetadata, *metadataHeaderRequired, *setRetryAfterHeader)
 	s.KubeClient = client
 	s.MetadataIP = *metadataIP
 	s.MetadataPort = *metadataPort
--- a/pkg/nmi/server/server.go
+++ b/pkg/nmi/server/server.go
@@ -49,6 +49,7 @@ type Server struct {
 	NodeName                           string
 	IPTableUpdateTimeIntervalInSeconds int
 	MICNamespace                       string
+	AllowedNamespaceToInstanceMetadata string
 	Initialized                        bool
 	BlockInstanceMetadata              bool
 	MetadataHeaderRequired             bool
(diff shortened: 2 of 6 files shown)


radu-catalina (approved):
lgtm

aramase (commented):
Hello 👋🏻  Thank you for the PR. 

We are planning to replace AAD Pod Identity with the new [Azure Workload Identity](https://github.com/Azure/azure-workload-identity). As the azure-workload-identity is now open-source, we're no longer adding any new features to this project. We will continue patching critical bugs and security vulnerabilities until further notice. Here is the disclaimer we have added: https://github.com/Azure/aad-pod-identity#aad-pod-identity.

Please let us know if you have any questions.

adriananeci:
@aramase we've already evaluated the `Azure Workload Identity`. While it looks promising and much better than `aad-pod-identity`, it still has 2 major limitations based on our use case:
- `Azure Workload Identity` doesn't support yet Managed Identities. This is the main reason we're using `aad-pod-identity`
- `aad-pod-identity` can also allow us to block access to instance metadata, endpoint that might expose sensitive information.

Because of the above reasons we still have to stay with `aad-pod-identity` for a while. Once `Azure Workload Identity` will offer features parity with `aad-pod-identity`, or at least support for `Managed Identities`, we will reconsider our setup.

May you reconsider adding this feature? It is critical to us in order to run Azure CSI drivers in overlay network. Or if you have other options, please let us know, they are welcomed.

aramase:
> May you reconsider adding this feature? It is critical to us in order to run Azure CSI drivers in overlay network. Or if you have other options, please let us know, they are welcomed.

Which Azure CSI drivers are you running? Are they running on host network? If the driver is running on host network, then NMI won't intercept any of the requests. The iptable rules NMI adds is only applicable for pods which are non host network. 

There might be a way to piggyback on the existing `AzurePodIdentityException` CRD without/ with minimal code changes which won't be a new feature. I'll post here once I get a chance to try it.

adriananeci:
Both AzureDisk and AzureFile. `csi-azuredisk-node` and `csi-azurefile-node` components  are currently running in host network, but we plan to move them in overlay network for security reasons. We are aware that NMI intercept requests only in overlay network, and this is why we're trying to extend the way access to instance metadata is allowed so that we are able to move all Azure CSI components into overlay network.

adriananeci:
@aramase thanks, looks like a simple `AzurePodIdentityException` object like
```
kind: AzurePodIdentityException
metadata:
  name: csi-azuredisk-node-exception
  namespace: kube-system
spec:
  podLabels:
    component: csi-azuredisk-node
```

did the trick. As a side note, only AzureDisk CSI needs access to instance metadata.

adriananeci:
Or not 😞 . Even if pods started fine, they still cannot access instance metadata:
```
W0128 22:46:25.994547       1 nodeserver.go:325] get instance type(vmss-agent-worker-test-spuil000000) failed with: failure of getting instance metadata with response "403 Forbidden"
W0128 22:46:25.995249       1 nodeserver.go:340] get zone(vmss-agent-worker-test-spuil000000) failed with: failure of getting instance metadata with response "403 Forbidden"
```

Also tried from inside on container:
```
/ # curl http://169.254.169.254/metadata/instance?api-version=2021-01-01
Request blocked by AAD Pod Identity NMI
```

aramase:
Thanks for trying it out with `AzurePodIdentityException`. I threw up a small change to make this work: https://github.com/Azure/aad-pod-identity/compare/master...aramase:block-instance-metadata?expand=1 but even this change introduces a new API server dependency in the instance metadata path. This change also might warrant a new minor release as it changes the default behavior which can be considered a _breaking change_.

IMO, `--block-instance-metadata` or proxying requests based on a policy for instance metadata requests is not something NMI should handle. The purpose of NMI is to use `AzureIdentity` and `AzureIdentityBinding` to validate and fetch tokens. Any other requests to IMDS shouldn't be the NMI's responsibility to validate/allow/deny. We already added `--block-instance-metadata` but I would prefer we don't overload this feature more and instead keep NMI only for the token requests.

adriananeci:
@aramase you're probably right about the purpose of NMI and the fact that it should not proxying requests based on a policy for instance metadata requests, but we're trying to migrate to Azure CSI drivers and avoid introducing other breaking changes on our end. Here the temporary solution to conditionally open access to instance metadata based on some sort of policies.

I don't think this is a real _breaking change_ since access to instance metadata is denied to everyone if `--block-instance-metadata` flag is set. By adding your changes, we are just enabling the flexibility to conditionally allow access to instance metadata based on an `AzurePodIdentityException`.

adriananeci:
@aramase any news on this initiative? I can adjust the PR based on your proposed changes if that is fine for you.

PR #1284, opened 2022-06-09
Allow selecting bind address for NMI

whites11:
**Reason for Change**:

When using NMI on a cluster using cilium as CNI, the iptables DNAT rule that intercepts traffic towards IMDS has to be replaced by a CiliumLocalRedirectPolicy that works using eBPF.

In practice, it means that traffic is redirected to `add-pod-identity` pod's IP (which is the node IP) rather than 127.0.0.1.

This currently fails, because NMI only listens to localhost.

See https://github.com/cilium/cilium/issues/20103 for more details

This PR adds a new flag to allow specifying which address port 2579 should listen to.

**Requirements**

- [x] squashed commits
- [ ] included documentation
- [x] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:

The default value of the new flag is 127.0.0.1 to make this change 100% compatible with previous releases.

### DIFF
--- a/pkg/nmi/server/server.go
+++ b/pkg/nmi/server/server.go
@@ -44,6 +44,7 @@ const (
 // the server. These can be set via command line.
 type Server struct {
 	KubeClient                         k8s.Client
+	NMIHost                            string
 	NMIPort                            string
 	MetadataIP                         string
 	MetadataPort                       string
@@ -104,17 +105,23 @@ func (s *Server) Run() error {
 	}
 	rtr.PathPrefix("/").HandlerFunc(s.defaultPathHandler)
 
-	klog.Infof("listening on port %s", s.NMIPort)
-	if err := http.ListenAndServe("localhost:"+s.NMIPort, rtr); err != nil {
+	klog.Infof("listening on %s:%s", s.NMIHost, s.NMIPort)
+	if err := http.ListenAndServe(fmt.Sprintf("%s:%s", s.NMIHost, s.NMIPort), rtr); err != nil {
 		klog.Fatalf("error creating http server: %+v", err)
 	}
 	return nil
 }
 
 func (s *Server) updateIPTableRulesInternal() {
-	klog.V(5).Infof("node(%s) ip(%s) metadata address(%s:%s) nmi port(%s)", s.NodeName, localhost, s.MetadataIP, s.MetadataPort, s.NMIPort)
+	target := s.NMIHost
+	if target == "0.0.0.0" {
+		// if we're binding to all interfaces, we still want to add iptables rules for localhost only
+		target = localhost
+	}
 
-	if err := iptables.AddCustomChain(s.MetadataIP, s.MetadataPort, localhost, s.NMIPort); err != nil {
+	klog.V(5).Infof("node(%s) ip(%s) metadata address(%s:%s) nmi port(%s)", s.NodeName, target, s.MetadataIP, s.MetadataPort, s.NMIPort)
+
+	if err := iptables.AddCustomChain(s.MetadataIP, s.MetadataPort, target, s.NMIPort); err != nil {
 		klog.Fatalf("%s", err)
 	}
 	if err := iptables.LogCustomChain(); err != nil {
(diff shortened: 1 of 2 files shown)


whites11:
Tests fail because of image vulnerability, doesn't feel like related to my proposed change.

aramase:
> Tests fail because of image vulnerability, doesn't feel like related to my proposed change.

The CVE will be patched with https://github.com/Azure/aad-pod-identity/pull/1285. Once that's merged, we'll need to rebase this PR.

pkg/nmi/server/server.go
```diff
@@ -104,8 +105,8 @@ func (s *Server) Run() error {
 	}
 	rtr.PathPrefix("/").HandlerFunc(s.defaultPathHandler)
 
-	klog.Infof("listening on port %s", s.NMIPort)
-	if err := http.ListenAndServe("localhost:"+s.NMIPort, rtr); err != nil {
+	klog.Infof("listening on %s:%s", s.NMIHost, s.NMIPort)
+	if err := http.ListenAndServe(fmt.Sprintf("%s:%s", s.NMIHost, s.NMIPort), rtr); err != nil {
```
aramase, line 109: The iptable rules need to use the host IP that NMI is listening on, so this needs to be updated too:
https://github.com/Azure/aad-pod-identity/blob/6cf9bac500d8f3f29492a6fd53f2e8990d51d9f0/pkg/nmi/server/server.go#L117-L119

aramase (commented):
@whites11 Thanks for the PR! Added a comment. 

Our test suite will only validate with `localhost` as that's the default option. Could you confirm if you have validated a custom build with a different IP that's not localhost?

pkg/nmi/server/server.go
whites11: Uhm good point. Lemme think about that, it might be a little tricky

whites11:
OK I made some changes and tested with all 3 scenarios:

Bind to all NICs:

```
$ kubectl -n giantswarm get po azure-ad-pod-identity-app-nmi-kkz8q -o yaml|grep nmi-listen
    - --nmi-listen-address=0.0.0.0

# iptables-save | grep 2579
-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:2579


# netstat -ntpl | grep 2579
tcp6       0      0 :::2579                 :::*                    LISTEN      504560/nmi
```

Bind to 127.0.0.1 (default)

```
$ kubectl -n giantswarm get po azure-ad-pod-identity-app-nmi-8rgvs -o yaml|grep nmi-listen
    - --nmi-listen-address=127.0.0.1
    
# iptables-save | grep 2579
-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:2579


# netstat -ntpl | grep 2579
tcp        0      0 127.0.0.1:2579          0.0.0.0:*               LISTEN      508739/nmi
```

Bind to a specified IP (doesn't make much sense as this IP should exist in all nodes or the pod will obviously fail to start)

```
$ kubectl -n giantswarm get po azure-ad-pod-identity-app-nmi-7zc8f -o yaml|grep nmi-listen
    - --nmi-listen-address=10.0.2.13
    
    
# iptables-save | grep 2579
-A aad-metadata ! -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DNAT --to-destination 10.0.2.13:2579


# netstat -ntpl | grep 2579
tcp        0      0 10.0.2.13:2579          0.0.0.0:*               LISTEN      775371/nmi   

```

WDYT?

aramase:
> OK I made some changes and tested with all 3 scenarios:

Yeah, that works!

whites11:
I don't udnerstand if the failing test is a flake or some error I introduced

whites11:
@aramase is there anything else I need to do to unblock this PR?

aramase (approved):

chewong (approved):

whites11:
Thank you @chewong @aramase for all the help ❤️

chewong:
@whites11 we will include this PR as part of the v1.8.10 release.

PR #1297, opened 2022-07-28
fix: set Content-Length header in NMI response

aramase:

**Reason for Change**:

If the Content-Length header is not explicitly set and the length is > 2048 bytes, writer.Write() method will automatically set the `Transfer-Encoding: Chunked`. To avoid this, we are setting the content length to the len of response when returning from NMI.

**Requirements**

- [ ] squashed commits
- [ ] included documentation
- [ ] ran `make precommit`

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


sozercan (approved):
LGTM

PR #1401, opened 2023-02-27
Update NMI server resource id query key

fenzhanG2:
**Reason for Change**:
In the latest .Net ["Azure.Identity" SDK](https://www.nuget.org/packages/Azure.Identity) (v1.8.2), the get identity token request to IMDS with resource id is like this:

Caller method example:
`var token = new ManagedIdentityCredential("<resourceId>").GetToken(<token request context>)`
HTTP request: _/metadata/identity/oauth2/token?api-version=...&resource=...&**mi_res_id**=<resource_id>_

So nmi needs to read both "mi_res_id" and "msi_res_id"(backward compatible) values.

Does this change contain code from or inspired by another project?  -  no

from Microsoft COSMIC team.


fenzhanG2:
@microsoft-github-policy-service agree company="Microsoft"

aramase (approved):

